PR #3, opened 2021-03-01
Extend the VolumeReplicationClass CRD

raghavendra-talur:


Madhu-1:
@leseb @jarrpa tagging you guys to help with the review

jarrpa:
Unfortunately, I have no experience with Kustomize. I imagine this came out of the box with the `operator-sdk init` command... I'm not sure ocs-operator used that, and if so it was a loooong time ago so things have definitely changed, and drastically so.

The only thing I can really say is that "Extend the VolumeReplicationClass CRD" as a PR title doesn't really capture what's going on, since you are basically initializing the entire project. I would suggest splitting this PR, having one to just lay down the boilerplate and then a follow-up to actually start in on the modifications. In either case, it would probably be good practice to make sure whatever you have builds and runs.

Madhu-1:
closing in favor of https://github.com/kube-storage/volume-replication-operator/pull/17

PR #2, opened 2021-03-01
Init the CRDs and the controllers

raghavendra-talur:
This commit is the only scaffolding required for the

* 3 CRDs 
** VolumeReplicationRequest
** VolumeReplica
**  VolumeReplicationClass

* 2 controllers
** volumereplicationrequest_controller
** volumereplica_controller

PR #12, opened 2021-03-02
Revert "Init the CRDs and the controllers"

raghavendra-talur:
Reverts kube-storage/volume-replication-operator#2

We decided to merge the request and the replica controllers into one controller that runs in the side car.


raghavendra-talur:
@Madhu-1 @umangachapagain @sp98 Please take a look.

Madhu-1 (approved):

PR #15, opened 2021-03-04
adds VolumeReplication CRD spec

sp98:


config/rbac/role.yaml
```diff
@@ -6,32 +6,6 @@ metadata:
   creationTimestamp: null
   name: manager-role
 rules:
-- apiGroups:
```
sp98, line 9: `make manifests` command deletes `role` config related to the volumeReplicationClass

api/v1alpha1/volumereplication_types.go
```diff
@@ -20,20 +20,77 @@ import (
...
+	MirroringEnabled ReplicationState = "MirroringEnabled"
+
+	// ImagePromoted means image was promoted to primary
+	ImagePromoted ReplicationState = "Promoted"
+
+	// ImageDemoted means image was demoted to secondary
+	ImageDemoted ReplicationState = "Demoted"
+
+	// ImageResynced means image was resynced
+	ImageResynced ReplicationState = "Resynced"
```
Madhu-1, line 51: for success can it be only 3 states. Primary,Secondary ,Resync

api/v1alpha1/volumereplication_types.go
```diff
@@ -20,20 +20,77 @@ import (
...
+
+	// ImageResyncFailed means image resync operation failed
+	ImageResyncFailed ReplicationState = "ResyncFailed"
+)
+
+// DataSourceSpec is
+type DataSourceSpec struct {
+	APIGroup string `json:"apiGroup"`
+	Kind     string `json:"kind"`
+	// Name is the PVC name associated with the image
```
Madhu-1, line 70: not related to pvc, name can be anything just the name of the object.

Madhu-1:
@travisn @leseb @jarrpa PTAL

api/v1alpha1/volumereplication_types.go
```diff
@@ -20,20 +20,77 @@ import (
...
+	ImagePromoted ReplicationState = "Promoted"
+
+	// ImageDemoted means image was demoted to secondary
+	ImageDemoted ReplicationState = "Demoted"
+
+	// ImageResynced means image was resynced
+	ImageResynced ReplicationState = "Resynced"
+
+	// MirroringFailed means mirroring operation was failed
+	MirroringFailed ReplicationState = "MirroringFailed"
```
leseb, line 54: Does this mean enabling mirroring failed?

api/v1alpha1/volumereplication_types.go
```diff
@@ -20,20 +20,77 @@ import (
...
+	// DataSource represents the PVC name associated with the image
+	DataSource DataSourceSpec `json:"dataSource"`
 }
 
 // VolumeReplicationStatus defines the observed state of VolumeReplication
 type VolumeReplicationStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	State          ReplicationState `json:"state,omitempty"`
+	Completed      bool             `json:"completed,omitempty"`
```
leseb, line 90: What is completed?

sp98: Just had a discussion with Madhu and Umanga. Will be removing this `Completed`  field.

api/v1alpha1/volumereplication_types.go
sp98: Yes, that was the original intent.  But now we are planning to keep it simple with following options only. 

```
// ReplicationState captures the latest state of the replication operation
type ReplicationState string

const (
	//Success signifies that the last replication operation succeeded
	Success ReplicationState = "Success"

	//Failure signifies that the last replication operation failed
	Failure ReplicationState = "Failure"
)
```
In case of failure, the details will be captured in the `Message` field if the `Status`. 

What do you think?

leseb: I think it's hard to have a "success" state for something (the replication) that runs continuously. For me, "success" is more a one-time action, which once completed can be reflected as a success.

I can see a state such as "Mirroring" or "Replicating".

The failure makes sense if for whatever reason the replication is stopped and data is not replicated anymore.

ShyamsundarR: Is there any intention to add status.conditions? Checking to see if we are elaborating any further conditions in here.

Also, A user can flip the `imageState` from primary(op-1)->secondary(op-2)->primary(op-3) even before the reconciler gets a chance to act on (or complete) the intermediate change (op-2). Thus a status that reflects primary as the result from op-1 spec would be stale as observed by the user. Are there plans to address this? (I would suggest adding to status an observed generation number, like [replicasets](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#replicasetstatus-v1-apps), but there maybe better patterns).

api/v1alpha1/volumereplication_types.go
```diff
@@ -20,20 +20,77 @@ import (
...
-	// Important: Run "make" to regenerate code after modifying this file
+	// VolumeReplicationClass is the VolumeReplicationClass name for this VolumeReplication resource
+	VolumeReplicationClass string `json:"volumeReplicationClass"`
+
+	// ImageState represents the replication operation to be performed on the image.
+	// Supported operations are "primary", "secondary" and "resync"
+	ImageState ImageState `json:"imageState"`
 
+	// DataSource represents the PVC name associated with the image
+	DataSource DataSourceSpec `json:"dataSource"`
```
ShyamsundarR, line 84: (asking for clarification) is there a reason not to use DataSource *corev1.TypedLocalObjectReference?

api/v1alpha1/volumereplication_types.go
umangachapagain: > Is there any intention to add status.conditions? Checking to see if we are elaborating any further conditions in here.
>
Yes. That's what I am proposing as a follow-up PR since there are some implementation details to be figured out.
>
> Also, A user can flip the `imageState` from primary(op-1)->secondary(op-2)->primary(op-3) even before the reconciler gets a chance to act on (or complete) the intermediate change (op-2). Thus a status that reflects primary as the result from op-1 spec would be stale as observed by the user. Are there plans to address this? (I would suggest adding to status an observed generation number, like [replicasets](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#replicasetstatus-v1-apps), but there maybe better patterns).
>
But how would observed generation help?
I wouldn't expect the users to check `Status.ObservedGeneration` before making the changes
so eventually when they do decide to check, they wouldn't know what the initial generation was and what should be the latest generation.
Or am I completely missing the point?

ShyamsundarR: > > Also, A user can flip the `imageState` from primary(op-1)->secondary(op-2)->primary(op-3) even before the reconciler gets a chance to act on (or complete) the intermediate change (op-2). Thus a status that reflects primary as the result from op-1 spec would be stale as observed by the user. Are there plans to address this? (I would suggest adding to status an observed generation number, like [replicasets](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#replicasetstatus-v1-apps), but there maybe better patterns).
> 
> But how would observed generation help?
> I wouldn't expect the users to check `Status.ObservedGeneration` before making the changes
> so eventually when they do decide to check, they wouldn't know what the initial generation was and what should be the latest generation.
> Or am I completely missing the point?

The user needs to know that the `observedGeneration` is the latest generation or the same generation as the CR is currently. That would be the only check a user can perform.

IOW, you are right, if `observedGeneration` is older than current generation, there is no way for a user to know what was desired in an older generation, and the user needs to wait till the generation catches up to know desired state was reached.

api/v1alpha1/volumereplication_types.go
```diff
@@ -20,20 +20,77 @@ import (
...
+
+const (
+	// Primary ImageState enables mirroring and promotes the image to primary
+	Primary ImageState = "primary"
+
+	// Secondary ImageState demotes the image to secondary and resyncs the image if out of sync
+	Secondary ImageState = "secondary"
+
+	// Resync option resyncs the image
+	Resync ImageState = "resync"
```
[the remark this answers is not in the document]
sp98: only `resync` action will be performed.

Madhu-1 (approved):

api/v1alpha1/volumereplication_types.go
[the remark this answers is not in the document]
ShyamsundarR: I am not sure this answers the question. Adding my thoughts here for the use-case, it does not prevent this PR from proceeding though.

- If Desired state is resync, the actual state should be secondary
- If Desired state is resync, and actual state is primary it would report an error in a specific status, as backend replication may still be happening (hence a specific status field or something)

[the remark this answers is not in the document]
Madhu-1: yes it can be implemented in the controller

* We can check the image replication state first
* If it is secondary we can resync
* If it's primary we can report an error.

It's the user's responsibility to keep the actual state in secondary before resyncing.

Madhu-1:
@ShyamsundarR @leseb Any other comments on this one?

ShyamsundarR (commented):
LGTM to start with the reconciler. I expect status and maybe spec to expand as the implementation proceeds. But a good start point.

PR #16, opened 2021-03-04
gactions: Add golangci-lint linter CI

iamniting:

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,0 +1,38 @@
+name: golangci-lint
+
+on:
+  pull_request:
+    branches: [ '*' ]
+    paths: ['**.go']
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v2
+      with:
+        # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+        version: v1.29
+
+        # Optional: working directory, useful for monorepos
+        # working-directory: somedir
+
+        # Optional: golangci-lint command line arguments.
+        # TODO: remove --skip-files controllers/suite_test.go once logic is implemented
+        args: -E gosec --timeout=6m --skip-files controllers/suite_test.go
+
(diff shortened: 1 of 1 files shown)


.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,38 @@
+name: golangci-lint
+
+on:
+  push:
```
Madhu-1: remove push?

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v2
+      with:
+        # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+        version: v1.29
+
+        # Optional: working directory, useful for monorepos
+        # working-directory: somedir
+
+        # Optional: golangci-lint command line arguments.
```
Madhu-1, line 23: add TODO to remove skip-files?

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,36 @@
+name: golangci-lint
+
+on:
+  pull_request:
+    branches: [ '*' ]
```
sp98, line 5: don't we only want it to run when PR is created for `main` branch ?

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,36 @@
+name: golangci-lint
+
+on:
+  pull_request:
```
sp98, line 4: How about adding `paths`. We can run the go lint only when PR involves changes in golang file. Suppose PR involves changes only in the README file, then we don't want the lint to run.  (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths)

.github/workflows/golangci-lint.yaml
iamniting: IMO we should run for all of them it will be useful at the time of backports.

.github/workflows/golangci-lint.yaml
iamniting: Done

sp98 (approved):

Madhu-1 (approved):

PR #17, opened 2021-03-04
add spec to VolumeReplicationClass CRD

umangachapagain:
Starting with minimal spec for VolumeReplicationClass and getting rid of dummy content.

Replacement for #3 


ShyamsundarR (commented):
As a start point this is enough.

As we expand the `ReplicationClass` we can look at adding storage specific parameters that denote, to the storage system, which storage end points are in the replication relationship etc. Most of these would get converted to parameters IMO and make the class immutable post creation like the `StorageClass`

sp98 (approved):

Madhu-1 (approved):

PR #19, opened 2021-03-05
RPC: add GRPC client wrappers to make replication call to csidriver

Madhu-1:
Added wrapper functions to make required calls to the csidriver to get the driver name and also to make the replication GRPC calls to the CSI driver.


pkg/replication/replication-client.go
```diff
@@ -0,0 +1,118 @@
...
+	// ResyncVolume RPC call to resync the volume.
+	ResyncVolume(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.
+		ResyncVolumeResponse, error)
+}
+
+func NewReplicationClient(cc *grpc.ClientConn, timeout time.Duration) VolumeReplication {
+	return &replicationClient{client: replicationlib.NewControllerClient(cc), timeout: timeout}
+}
+
+func (rc *replicationClient) EnableVolumeReplication(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.EnableVolumeReplicationResponse, error) {
```
sp98, line 52: missing GoDoc like rest of the methods.

sp98 (changes requested):
nit

pkg/replication/replication-client.go
Madhu-1: added now.

sp98 (approved):

pkg/replication/replication-client.go
```diff
@@ -0,0 +1,122 @@
...
+	DemoteVolume(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.
+		DemoteVolumeResponse, error)
+	// ResyncVolume RPC call to resync the volume.
+	ResyncVolume(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.
+		ResyncVolumeResponse, error)
+}
+
+// NewReplicationClient returns VolumeReplication interface which has the RPC
+// calls for replication.
+func NewReplicationClient(cc *grpc.ClientConn, timeout time.Duration) VolumeReplication {
```
ShyamsundarR, line 51: If this is per volume (which would be short lived as a volume is handled in one iteration of the reconcile function) we can possibly enhance the New by adding the volID and secrets, such that other calls are not passing this in repetitively.

If this is once per ReplicationClass, we may not want to retain secrets in here for the controller lifetime.

These can be improved later, as it stands it makes code more elegant rather than creating structures everywhere before calling the gRPC.

ShyamsundarR (commented):
LGTM! We can deal with interface betterment over time (or as other pieces use them) IMO.

pkg/replication/replication-client.go
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright 2021.
```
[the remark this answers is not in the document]
Madhu-1: Not added kube-storage authors to any files as of now in this repo.

pkg/replication/replication-client.go
Madhu-1: good for enhancements  :+1:  will keep this in mind.

umangachapagain (approved):

iamniting (approved):

PR #21, opened 2021-03-08
Add new driver specific configurations

Madhu-1:
This PR does below changes

* Add New DriverConfig structure which holds the required information to establish the connection with CSI driver
* dynamic generation of  leader election ID 
* Add leader election Namespace
* Change health-probe-bind-address to `9998`
* Disable Metrics by default

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -25,13 +25,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	replicationv1alpha1 "github.com/kube-storage/volume-replication-operator/api/v1alpha1"
+	"github.com/kube-storage/volume-replication-operator/pkg/config"
 )
 
 // VolumeReplicationReconciler reconciles a VolumeReplication object
 type VolumeReplicationReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Log          logr.Logger
+	Scheme       *runtime.Scheme
+	DriverConfig *config.DriverConfig
 }
 
 // +kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumereplications,verbs=get;list;watch;create;update;patch;delete
@@ -56,7 +58,8 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.DriverConfig) error {
+	r.DriverConfig = cfg
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&replicationv1alpha1.VolumeReplication{}).
 		Complete(r)
--- a/main.go
+++ b/main.go
@@ -52,7 +52,7 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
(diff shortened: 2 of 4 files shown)


iamniting (approved):

controllers/volumereplication_controller.go
```diff
@@ -25,13 +25,15 @@ import (
...
 	replicationv1alpha1 "github.com/kube-storage/volume-replication-operator/api/v1alpha1"
+	"github.com/kube-storage/volume-replication-operator/pkg/config"
 )
 
 // VolumeReplicationReconciler reconciles a VolumeReplication object
 type VolumeReplicationReconciler struct {
 	client.Client
 	Log    logr.Logger
 	Scheme *runtime.Scheme
+	Config *config.DriverConfig
```
umangachapagain, line 36:
```suggestion
	DriverConfig *config.DriverConfig
```
to be clearer?

main.go
```diff
@@ -50,10 +58,18 @@ func init() {
...
 	var probeAddr string
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+
+	cfg := config.NewDriverConfig()
+
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metric endpoint binds to.")
+	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "", "Namespace where the leader election resource lives")
+	flag.StringVar(&cfg.DriverName, "driver-name", "", "The CSI driver name.")
+	flag.StringVar(&cfg.DriverEndPoint, "csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
```
umangachapagain, line 70:
```suggestion
	flag.StringVar(&cfg.DriverEndpoint, "csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
```

main.go
```diff
@@ -50,10 +58,18 @@ func init() {
...
+	flag.DurationVar(&cfg.Timeout, "timeout", defaultTimeout, "The timeout for RPCs to the CSI driver.")
```
umangachapagain, line 71:
```suggestion
	flag.DurationVar(&cfg.Timeout, "rpc-timeout", defaultTimeout, "The timeout for RPCs to the CSI driver.")
```

main.go
```diff
@@ -50,10 +58,18 @@ func init() {
...
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":9998", "The address the probe endpoint binds to.")
```
umangachapagain, line 72: Should we keep this as before and change it in our deployment configuration? Or do you want to enforce this default to everyone (unless they want to change it) ?

main.go
```diff
@@ -65,13 +81,28 @@ func main() {
...
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	err := cfg.Validate()
+	if err != nil {
+		setupLog.Error(err, "error in driver configuration")
+		os.Exit(1)
+	}
+
+	if leaderElectionNamespace == "" {
+		fmt.Println("leader-election-namespace is empty")
```
umangachapagain, line 91: Should/Can this be logged as well?

pkg/config/config.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"errors"
+	"time"
+)
+
+// DriverConfig holds the CSI driver configuration.
+type DriverConfig struct {
+	// The gRPC endpoint for CSI driver.
+	DriverEndPoint string
```
umangachapagain, line 27:
```suggestion
	DriverEndpoint string
```

pkg/config/config.go
```diff
@@ -0,0 +1,46 @@
...
+	// The name of CSI Driver.
+	DriverName string
+	// The timeout for RPCs to the CSI driver.
+	Timeout time.Duration
```
umangachapagain, line 31:
```suggestion
	RPCTimeout time.Duration
```

sp98 (commented):
nit. 
rest LGTM

sp98 (approved):

main.go
Madhu-1: changed it 9998 as 8081 is the most common and widely used one.

main.go
Madhu-1: I tried to log initially it becomes messy as we need to define a new error variable and also `setupLog.Error` Needs two arguments.

main.go
umangachapagain: I had some other thoughts, but I this is fine too.

main.go
umangachapagain: yes, but with this approach we lose log when we exit. is there a way to debug why we exited?
If yes, then I won't insist on logging.

Madhu-1: done now logging to stderr

umangachapagain (approved):

PR #22, opened 2021-03-08
Add unit testing for replication client

Madhu-1:
Added Mock client for replication and also the unit testing for the same.

### DIFF
--- a/pkg/client/fake/replication-client.go
+++ b/pkg/client/fake/replication-client.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fake
+
+import (
+	"context"
+
+	replicationlib "github.com/kube-storage/spec/lib/go/replication"
+)
+
+// ReplicationClient to fake replication operations.
+type ReplicationClient struct {
+	// EnableVolumeReplicationMock mocks EnableVolumeReplication RPC call.
+	EnableVolumeReplicationMock func(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.EnableVolumeReplicationResponse, error)
+	// DisableVolumeReplicationMock mocks DisableVolumeReplication RPC call.
+	DisableVolumeReplicationMock func(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.DisableVolumeReplicationResponse, error)
+	// PromoteVolumeMock mocks PromoteVolume RPC call.
+	PromoteVolumeMock func(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.PromoteVolumeResponse, error)
+	// DemoteVolumeMock mocks DemoteVolume RPC call.
+	DemoteVolumeMock func(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.DemoteVolumeResponse, error)
+	// ResyncVolumeMock mocks ResyncVolume RPC call.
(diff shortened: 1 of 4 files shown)


pkg/replication/replication-client_test.go
```diff
@@ -0,0 +1,153 @@
...
+	// return success response
+	mockedEnableReplication := &replicationClientMock{
+		EnableVolumeReplicationMock: func(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.EnableVolumeReplicationResponse, error) {
+			return &replicationlib.EnableVolumeReplicationResponse{}, nil
+		},
+	}
+	client = mockedEnableReplication
+	resp, err := client.EnableVolumeReplication(context.Background(), "", nil, nil)
+	assert.Equal(t, &replicationlib.EnableVolumeReplicationResponse{}, resp)
+	assert.Nil(t, err)
```
umangachapagain, line 41:
```suggestion
	// return success response
	mockedEnableReplication := &replicationClientMock{
		EnableVolumeReplicationMock: func(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.EnableVolumeReplicationResponse, error) {
			return &replicationlib.EnableVolumeReplicationResponse{}, nil
		},
	}
	resp, err := mockedEnableReplication.EnableVolumeReplicationMock(context.Background(), "", nil, nil)
	assert.Equal(t, &replicationlib.EnableVolumeReplicationResponse{}, resp)
	assert.Nil(t, err)
```

I tried this and it works too. Is there a reason why you first do `client = NewReplicationClient()` and then overwrite it with `mockedEnableReplication` ?

umangachapagain:
Code looks good, but I am a unsure about the overall structure and naming.
Can we have
```
|-pkg/
|--client/
|----client.go
|----replication.go
|----fake/
|------replication.go
```
That way all clients are at one place and fake(or mocks) are at one place.
Also when you `import client/fake`, you could do `fake.replicationClient` and when you `import client/replication`, you can do `replication.replicationClient`

pkg/replication/replication-client_test.go
Madhu-1: Yes, it would work but from the Reconcile logic, we cannot call mockedEnableReplication.EnableVolumeReplicationMock directly . it will call `client.EnableVolumeReplication` as we are assigning `client = mockedEnableReplication` it will ensure that when the user calls the actual GRPC client. the mock functions we are writing will get executed.

umangachapagain: aah makes sense. Thanks.

umangachapagain (approved):

pkg/client/fake/replication-client.go
```diff
@@ -0,0 +1,62 @@
...
+	"context"
+
+	replicationlib "github.com/kube-storage/spec/lib/go/replication"
+)
+
+// ReplicationClient to fake replication operations.
+type ReplicationClient struct {
+	// EnableVolumeReplicationMock mocks EnableVolumeReplication RPC call.
+	EnableVolumeReplicationMock func(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.EnableVolumeReplicationResponse, error)
+	// DisableVolumeReplicationMock  mocks DisableVolumeReplication RPC call.
```
sp98, line 29:
```suggestion
	// DisableVolumeReplicationMock mocks DisableVolumeReplication RPC call.
```

pkg/client/fake/replication-client.go
```diff
@@ -0,0 +1,62 @@
...
+func (rc *ReplicationClient) EnableVolumeReplication(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.EnableVolumeReplicationResponse, error) {
+	return rc.EnableVolumeReplicationMock(ctx, volumeID, secrets, parameters)
+}
+
+// DisableVolumeReplication calls DisableVolumeReplicationMock mock function.
+func (rc *ReplicationClient) DisableVolumeReplication(ctx context.Context, volumeID string, secrets, parameters map[string]string) (*replicationlib.DisableVolumeReplicationResponse, error) {
+	return rc.DisableVolumeReplicationMock(ctx, volumeID, secrets, parameters)
+}
+
+// PromoteVolume calls  PromoteVolumeMock mock function.
```
sp98, line 49:
```suggestion
// PromoteVolume calls PromoteVolumeMock mock function.
```

sp98 (commented):
nit.

Madhu-1:
@sp98 addressed the nit.

sp98 (approved):

PR #33, opened 2021-03-11
gactions: Run tests and make build

iamniting:
Run `make docker-build` it will run `make test` as well. This way we can
run tests and make sure build are working. It will make sure that we can
create buids after merging a PR.

### DIFF
--- a/.github/workflows/test-and-build.yaml
+++ b/.github/workflows/test-and-build.yaml
@@ -1,4 +1,4 @@
-name: make test
+name: docker-build
 
 on:
   pull_request:
@@ -6,7 +6,7 @@ on:
 
 jobs:
   test:
-    name: Test
+    name: test and build
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-build.yaml
```diff
@@ -1,12 +1,12 @@
-name: make test
+name: make test and docker-build
```
Madhu-1, line 1: as we are running only docker build change this do make docker-build and also jobs name need to be updated

iamniting: Actually this will run test also as part of build 

https://github.com/kube-storage/volume-replication-operator/blob/main/Makefile#L94

Madhu-1: Agree but that is not the action command we ran right? it might be doing it internally.

iamniting: Did it in a different way

Madhu-1 (approved):

PR #41, opened 2021-03-12
enable forced volume promotion in case of known errors

umangachapagain:

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -40,6 +42,7 @@ const (
 	enableVolumeReplication  = "Enable volume replication"
 	disableVolumeReplication = "Disable volume replication"
 	promoteVolume            = "Promote volume"
+	forcePromoteVolume       = "Promote volume (forced)"
 	demoteVolume             = "Demote volume"
 	resyncVolume             = "Resync volume"
 
@@ -160,7 +163,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	switch instance.Spec.ImageState {
 	case replicationv1alpha1.Primary:
-		err := r.markVolumeAsPrimary(volumeHandle, parameters, secret)
+		err := r.markVolumeAsPrimary(instance, volumeHandle, parameters, secret)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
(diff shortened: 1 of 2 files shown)


controllers/volumereplication_controller.go
```diff
@@ -209,26 +211,41 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+			Task: replication.NewPromoteVolumeTask(c, forcePromote),
+			KnownErrors: []codes.Code{
+				codes.FailedPrecondition,
+			},
 		},
 	}
 
-	return tasks.RunAll(markVolumeAsPrimaryTasks)
+	taskName, taskError := tasks.RunAll(markVolumeAsPrimaryTasks)
+	if taskError.Position >= 0 {
```
Madhu-1, line 236: why we need a Position here?

controllers/volumereplication_controller.go
```diff
@@ -209,26 +211,41 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+		s, ok := status.FromError(taskError.Message)
+		if !ok {
+			return taskName, taskError.Message
+		}
+		for _, e := range markVolumeAsPrimaryTasks[taskError.Position].KnownErrors {
+			if s.Code() == e {
+				r.markVolumeAsPrimary(volumeID, true, parameters, secrets)
```
Madhu-1, line 243: This will call the EnableReplication again. we can avoid one more network call if we take this approach https://github.com/Madhu-1/operator/commit/6ab7b66c67b78c3ceffd31b4fb8c70e01665bfac

controllers/volumereplication_controller.go
umangachapagain: Because we need to know which Task failed to be able to fetch it's KnownErrors.

controllers/volumereplication_controller.go
umangachapagain: Oh I forgot that we're dealing with GRPC calls. I'll fix it.

controllers/volumereplication_controller.go
Madhu-1: Take a look at https://github.com/kube-storage/volume-replication-operator/pull/42 based on the name we can check  the errors for the particular task

umangachapagain: That looks clean. Will rebase on top of it.

umangachapagain:
@Madhu-1 Made the changes as per your suggestions. But, I see 2 main issues with this implementation
1. In case we have `KnownErrors` we are adding it to the `taskResp` list. This means `markVolumeAsPrimary` will be considered as failed even if it didn't actually fail. If we don't add it to `KnownErrors`, we won't be able to log it and user won't know about it.
2. In case `PromoteVolume` passes without any errors, we move on to the next task which is `Force Promote`. This would be an unnecessary call that we need to avoid.

Madhu-1:
> @Madhu-1 Made the changes as per your suggestions. But, I see 2 main issues with this implementation

Yes, correct we need to solve it.

> 
> 1. In case we have `KnownErrors` we are adding it to the `taskResp` list. This means `markVolumeAsPrimary` will be considered as failed even if it didn't actually fail. If we don't add it to `KnownErrors`, we won't be able to log it and user won't know about it.

if we move the logic to `markVolumeAsPrimary` we can log it?
> 2. In case `PromoteVolume` passes without any errors, we move on to the next task which is `Force Promote`. This would be an unnecessary call that we need to avoid.

if we move the logic for error handling to `markVolumeAsPrimary` call force promotion only if the promotion failed with know error?

umangachapagain:
> > @Madhu-1 Made the changes as per your suggestions. But, I see 2 main issues with this implementation
> 
> Yes, correct we need to solve it.
> 
> > 1. In case we have `KnownErrors` we are adding it to the `taskResp` list. This means `markVolumeAsPrimary` will be considered as failed even if it didn't actually fail. If we don't add it to `KnownErrors`, we won't be able to log it and user won't know about it.
> 
> if we move the logic to `markVolumeAsPrimary` we can log it?
> 
We already log it from there.  But after logging we return with error. So we need to "not" return in case of knownErrors. And to do that we need to know if it is a known error.

> > 1. In case `PromoteVolume` passes without any errors, we move on to the next task which is `Force Promote`. This would be an unnecessary call that we need to avoid.
> 
> if we move the logic for error handling to `markVolumeAsPrimary` call force promotion only if the promotion failed with know error?
>
We do error handling there. But that would be too late since RunAll would have already run ForcePromote by the time we can handle it. We need to be able to dynamically add tasks to the list based on error from previous task,

Madhu-1:
somthing like below can be done?
```go
c := replication.CommonRequestParameters{
		VolumeID:    volumeID,
		Parameters:  parameters,
		Secrets:     secrets,
		Replication: r.Replication,
	}
	force := false
var markVolumeAsPrimaryTasks = []*tasks.TaskSpec{
		{
			Name: enableVolumeReplication,
			Task: replication.NewEnableTask(c),
		},
	}

resp := tasks.RunAll(markVolumeAsPrimaryTasks)
	// Check error for all tasks and return error
	for _, re := range resp {
		if re.Error != nil {
			r.Log.Error(re.Error, "task failed", "taskName", re.Name)
			return re.Error
		}
	}

// TODO check if the CSI driver supports promote volume
// if promote volume is not supported skip below tasks

var promoteVolumeTask = []*tasks.TaskSpec{
		{
			Name: promoteVolume,
			Task: replication.NewPromoteVolumeTask(c, force),
			KnownErrors: []codes.Code{
				codes.FailedPrecondition,
			},
		},

	}

for _, re := range resp {
		if re.Error != nil {
// check the error is known error if yes go for next task or return error
		}
	}

var forcePromoteTask = []*tasks.TaskSpec{
		{
		{
			Name: forcePromoteVolume,
			Task: replication.NewPromoteVolumeTask(c, true),
		},
	}

for _, re := range resp {
		if re.Error != nil {
// return error, If promotion failed add it to queue to know that mirroring is done only promotion is pending for this volume
		}
	}
```

Madhu-1:
If the task is making things complicated we can remove it and directly call the replication functions.

umangachapagain:
@Madhu-1 Made changes. Now we need to differentiate between Promote and Forced Promote. Other than that it should work.

controllers/volumereplication_controller.go
```diff
@@ -205,33 +208,97 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
 		}
 	}
+	instance.Status.State = replicationv1alpha1.Replicating
+	err = r.Client.Status().Update(context.TODO(), &instance, nil)
+	if err != nil {
+		return err
+	}
+
+	if instance.Spec.ImageState == replicationv1alpha1.Primary &&
+		instance.Status.State == replicationv1alpha1.Replicating {
```
iamniting, line 247: This will be always `replicating` as you are changing the `state` at line 240. Maybe we can remove this if statement.

controllers/volumereplication_controller.go
```diff
@@ -205,33 +208,97 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
-		{
-			Name: enableVolumeReplication,
-			Task: replication.NewEnableTask(c),
-		},
-		{
-			Name: promoteVolume,
-			Task: replication.NewPromoteVolumeTask(c, force),
-		},
+
+	instance := replicationv1alpha1.VolumeReplication{}
```
iamniting, line 219: Same as https://github.com/kube-storage/volume-replication-operator/pull/45#discussion_r595108903

iamniting (changes requested):

controllers/volumereplication_controller.go
```diff
@@ -205,33 +208,97 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+	err := r.Client.Get(context.TODO(), volumeReplicationObjectRef, &instance)
```
Madhu-1, line 220: do we need to get the instance again? just use the existing instance we already got?

controllers/volumereplication_controller.go
```diff
@@ -205,33 +208,97 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+				}
+				// if error is not a known error then we return
+				// else we continue on to next task
+				if !isKnownError {
+					r.Log.Error(re.Error, "task failed", "taskName", re.Name)
+					return re.Error
+				}
+			}
+		}
+
```
Madhu-1, line 284: move this to a helper function to check known errors. this can be reused later

controllers/volumereplication_controller.go
umangachapagain: Using existing instance.

Madhu-1 (approved):

Madhu-1:
@sp98 PTAL

iamniting (approved):

PR #42, opened 2021-03-15
task: Return error and response from the task

Madhu-1:
when we are all the tasks from the controller, the controller has to take few decisions based on the response received from the particular tasks. For example, the controller needs to add the Resync or Secondary volume request to the queue so that it is to wait till the resync is completed and the volume is to be marked as ready to use by application pods.

### DIFF
--- a/controllers/tasks/task.go
+++ b/controllers/tasks/task.go
@@ -23,19 +23,34 @@ type TaskSpec struct {
 	KnownErrors []error
 }
 
+// TaskResponse represents the response of each task
+type TaskResponse struct {
+	Name     string
+	Response interface{}
+	Error    error
+}
+
 // Task is a specific task to be done by controller
 type Task interface {
-	Run() error
+	Run() (interface{}, error)
 }
 
 // RunAll executes all the Task in the given list of TaskSpec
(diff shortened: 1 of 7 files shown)


iamniting (approved):

controllers/tasks/task.go
```diff
@@ -23,19 +23,34 @@ type TaskSpec struct {
...
 // Task is a specific task to be done by controller
 type Task interface {
-	Run() error
+	Run() (interface{}, error)
 }
 
 // RunAll executes all the Task in the given list of TaskSpec
-func RunAll(tasks []*TaskSpec) (string, error) {
+func RunAll(tasks []*TaskSpec) []*TaskResponse {
+	taskResp := []*TaskResponse{}
```
umangachapagain, line 40: Can we avoid pointer?

Madhu-1: Any specific reason? I used a pointer to avoid the high size of structs in return.

umangachapagain (approved):

controllers/tasks/task.go
umangachapagain: I was worried about heap allocations.

PR #43, opened 2021-03-15
vr: Update status of the volumereplication object

iamniting:
Update status of volume replication CR

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -91,6 +92,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	// Get VolumeReplicationClass
 	vrcObj, err := r.getVolumeReplicaCLass(instance.Spec.VolumeReplicationClass)
 	if err != nil {
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
 		return ctrl.Result{}, err
 	}
 
@@ -100,6 +102,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	err = validatePrefixedParameters(vrcObj.Spec.Parameters)
 	if err != nil {
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
 		r.Log.Error(err, "failed to validate prefix parameters", "volumeReplicationClass", instance.Spec.VolumeReplicationClass)
 		return ctrl.Result{}, err
 	}
@@ -113,6 +116,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	if secretName != "" && secretNamespace != "" {
 		secret, err = r.getSecret(secretName, secretNamespace)
 		if err != nil {
+			_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
 			return reconcile.Result{}, err
 		}
 	}
@@ -123,11 +127,13 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	case pvcDataSource:
 		_, pv, err := r.getPVCDataSource(nameSpacedName)
 		if err != nil {
+			_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
 			r.Log.Error(err, "failed to get dataSource for PVC", "dataSourceName", nameSpacedName.Name)
 			return ctrl.Result{}, err
 		}
 		volumeHandle = pv.Spec.CSI.VolumeHandle
 	default:
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, ("unsupported datasource kind " + instance.Spec.DataSource.Kind))
 		return ctrl.Result{}, fmt.Errorf("unsupported datasource kind %q", instance.Spec.DataSource.Kind)
 	}
 
@@ -162,36 +168,58 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
-	switch instance.Spec.ReplicationState {
-	case replicationv1alpha1.Primary:
-		err := r.markVolumeAsPrimary(instance, volumeHandle, parameters, secret)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-
-	case replicationv1alpha1.Secondary:
-		requeueForResync, err := r.markVolumeAsSecondary(instance, volumeHandle, parameters, secret)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		if requeueForResync {
-			return ctrl.Result{Requeue: true}, nil
-		}
-
-	case replicationv1alpha1.Resync:
-		requeueForResync, err := r.resyncVolume(instance, volumeHandle, parameters, secret)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		if requeueForResync {
-			return ctrl.Result{Requeue: true}, nil
-		}
-
-	default:
-		return ctrl.Result{}, fmt.Errorf("unsupported replication state %q", instance.Spec.ReplicationState)
+	instance.Status.LastStartTime = metav1.Now()
+	instance.Status.LastCompletionTime.Reset()
+	if err = r.Client.Update(context.TODO(), instance); err != nil {
+		r.Log.Error(err, "Failed to update status", "Object", instance.Name)
+		return reconcile.Result{}, err
 	}
 
-	return ctrl.Result{}, nil
+	var replicationErr error
+	var requeueForResync bool
+
+	switch instance.Spec.ReplicationState {
+	case replicationv1alpha1.Primary:
+		replicationErr = r.markVolumeAsPrimary(instance, volumeHandle, parameters, secret)
+
+	case replicationv1alpha1.Secondary:
+		requeueForResync, replicationErr = r.markVolumeAsSecondary(instance, volumeHandle, parameters, secret)
+
+	case replicationv1alpha1.Resync:
+		requeueForResync, replicationErr = r.resyncVolume(instance, volumeHandle, parameters, secret)
+
+	default:
+		replicationErr = fmt.Errorf("unsupported image state %q", instance.Spec.ReplicationState)
+	}
+
+	if replicationErr != nil {
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, replicationErr.Error())
+		return ctrl.Result{}, replicationErr
+	}
+
+	if requeueForResync {
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	instance.Status.LastCompletionTime = metav1.Now()
+	if instance.Spec.ReplicationState == replicationv1alpha1.Resync {
+		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked for resyncing")
+	} else {
+		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked "+string(instance.Spec.ReplicationState))
+	}
+
+	return ctrl.Result{}, err
+}
+
+func (r *VolumeReplicationReconciler) updateReplicationStatus(instance *replicationv1alpha1.VolumeReplication, state replicationv1alpha1.State, message string) error {
+	instance.Status.State = state
+	instance.Status.Message = message
+	if err := r.Client.Update(context.TODO(), instance); err != nil {
+		r.Log.Error(err, "Failed to update status", "Object", instance.Name)
+		return err
+	}
+
+	return nil
 }
 
 // SetupWithManager sets up the controller with the Manager.


api/v1alpha1/volumereplication_types.go
```diff
@@ -44,6 +44,9 @@ const (
 
 	// ReplicationFailure means the last replication operation failed
 	ReplicationFailure ReplicationState = "Failed"
+
+	// Replicated means the last replication operation is succeeded
+	Replicated ReplicationState = "Replicated"
```
Madhu-1, line 49: what is the diff b/w replicating and replicated here?

controllers/volumereplication_controller.go
```diff
@@ -163,31 +170,44 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 
+	r.updateStatus(instance, replicationv1alpha1.Replicated, "")
 	return ctrl.Result{}, nil
 }
 
+func (r *VolumeReplicationReconciler) updateStatus(instance *replicationv1alpha1.VolumeReplication, state replicationv1alpha1.ReplicationState, message string) {
+	instance.Status.State = state
+	instance.Status.Message = message
+	if err := r.Client.Update(context.TODO(), instance); err != nil {
+		r.Log.Info("Update Error", "MetaUpdateErr", "Failed to update status")
```
Madhu-1, line 207: error is not logged here?

Madhu-1 (changes requested):
@iamniting can you be more specific from what state we are changing to what each operation, a table to cover all operations like (primary, resync, and secondary) might be helpful to understand it.

api/v1alpha1/volumereplication_types.go
iamniting: replicating means it is in progress and replicated means replication is succeeded

controllers/volumereplication_controller.go
iamniting: Done

api/v1alpha1/volumereplication_types.go
Madhu-1: Replicated does not sound correct.  Replication will never go to the finish state right? from our side we have enabled the replication in the background the replication just now started and it's replicating till we disable it.

iamniting:
> @iamniting can you be more specific from what state we are changing to what each operation, a table to cover all operations like (primary, resync, and secondary) might be helpful to understand it.

Maybe then we can add a phase also to the status to differentiate b/w actual state and the reconcile state to clear all this doubts.

Madhu-1:
> > @iamniting can you be more specific from what state we are changing to what each operation, a table to cover all operations like (primary, resync, and secondary) might be helpful to understand it.
> 
> Maybe then we can add a phase also to the status to differentiate b/w actual state and the reconcile state to clear all these doubts.

adding this table for more clarity helps https://github.com/kube-storage/volume-replication-operator/pull/43#pullrequestreview-611885593

iamniting:
> > > @iamniting can you be more specific from what state we are changing to what each operation, a table to cover all operations like (primary, resync, and secondary) might be helpful to understand it.
> > 
> > 
> > Maybe then we can add a phase also to the status to differentiate b/w actual state and the reconcile state to clear all these doubts.
> 
> adding this table for more clarity helps [#43 (review)](https://github.com/kube-storage/volume-replication-operator/pull/43#pullrequestreview-611885593)


| ImageState | StatusPhase | StatusState |
| --------------- | --------------- | --------------- |
| Primary | Replicating/Failed/Replicated | Unknown/PrevState/Primary
| Secondary | Replicating/Failed/Replicated | Unknown/PrevState/Secondary
| Resync | Replicating/Failed/Replicated | Unknown/PrevState/Resynced

The unknown state will be for the newly created CR.
PrevState will be for the edited CR it will reflect PrevState until a new state is acheived.
Will mark the state as ImageState if no failures and achieved the state.

api/v1alpha1/volumereplication_types.go
iamniting: removed it.

controllers/volumereplication_controller.go
```diff
@@ -84,9 +85,13 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return reconcile.Result{}, err
 	}
 
+	instance.Status.LastStartTime = metav1.Now()
+	r.updateStatus(instance, replicationv1alpha1.Replicating, "")
```
Madhu-1, line 89: This should be set only in case of success?

controllers/volumereplication_controller.go
```diff
@@ -162,28 +171,42 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 			return ctrl.Result{}, err
 		}
 
 	default:
+		r.updateStatus(instance, replicationv1alpha1.ReplicationFailure, ("unsupported image state " + string(instance.Spec.ImageState)))
 		return ctrl.Result{}, fmt.Errorf("unsupported image state %q", instance.Spec.ImageState)
 	}
 
+	instance.Status.LastCompletionTime = metav1.Now()
+	r.updateStatus(instance, replicationv1alpha1.Replicating, "volume replication is ready to use")
```
Madhu-1, line 198: let's remove `volume replication is ready to use now. keep the message as `Volume is marked primary or {secondary}` in case of resync `volume is marked for resyncing  

@ShyamsundarR any suggestion here?

controllers/volumereplication_controller.go
iamniting: Done

controllers/volumereplication_controller.go
```diff
@@ -162,28 +171,42 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	return ctrl.Result{}, nil
 }
 
+func (r *VolumeReplicationReconciler) updateStatus(instance *replicationv1alpha1.VolumeReplication, state replicationv1alpha1.ReplicationState, message string) {
+	instance.Status.State = state
+	instance.Status.Message = message
+	if err := r.Client.Update(context.TODO(), instance); err != nil {
+		r.Log.Error(err, "Update Error", "MetaUpdateErr", "Failed to update status")
```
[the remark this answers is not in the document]
iamniting: The object status is metadata right that's why I wrote like this. even in the ocs-operator we do the same, otherwise, we can change this to the `ObjectUpdateErr` or something like that.

controllers/volumereplication_controller.go

controllers/volumereplication_controller.go
```diff
@@ -84,9 +85,13 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return reconcile.Result{}, err
 	}
 
+	instance.Status.LastStartTime = metav1.Now()
+	r.updateStatus(instance, replicationv1alpha1.ReplicationState(""), "")
```
sp98, line 89: I didn't understand this `ReplicationState(" ")` message call.  is this a way for returning empty ReplicationState type?

iamniting: yes

Madhu-1: Yes why this is set?

sp98: Also, what happens when the resource is being deleted?  Do we still want to perform this particular update?

controllers/volumereplication_controller.go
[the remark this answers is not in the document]
Madhu-1: can this be just `r.Log.Error(err, "Update Error", "Failed to update status")`

controllers/volumereplication_controller.go
```diff
@@ -162,28 +171,46 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		r.updateStatus(instance, replicationv1alpha1.Replicating, "volume is marked "+string(instance.Spec.ImageState))
+	}
 	return ctrl.Result{}, nil
 }
 
+func (r *VolumeReplicationReconciler) updateStatus(instance *replicationv1alpha1.VolumeReplication, state replicationv1alpha1.ReplicationState, message string) {
+	instance.Status.State = state
+	instance.Status.Message = message
+	if err := r.Client.Update(context.TODO(), instance); err != nil {
+		r.Log.Error(err, "Update Error", "MetaUpdateErr", "Failed to update status")
```
umangachapagain, line 210:
```suggestion
		r.Log.Error(err, "Failed to update status", "Object", instance )
```

controllers/volumereplication_controller.go
```diff
@@ -162,28 +171,46 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	case replicationv1alpha1.Resync:
 		err := r.resyncVolume(volumeHandle, parameters, secret)
 		if err != nil {
+			r.updateStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
 			return ctrl.Result{}, err
 		}
 
 	default:
+		r.updateStatus(instance, replicationv1alpha1.ReplicationFailure, ("unsupported image state " + string(instance.Spec.ImageState)))
 		return ctrl.Result{}, fmt.Errorf("unsupported image state %q", instance.Spec.ImageState)
```
umangachapagain, line 194: Define a new error and reuse it at both places instead of duplicating.

umangachapagain (changes requested):

controllers/volumereplication_controller.go

controllers/volumereplication_controller.go

controllers/volumereplication_controller.go
sp98: `instance.Name` ? @umangachapagain

controllers/volumereplication_controller.go
iamniting: Removed

controllers/volumereplication_controller.go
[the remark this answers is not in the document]
iamniting: Updated in diff way

controllers/volumereplication_controller.go
```diff
@@ -162,26 +174,49 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		err = r.updateStatusStateAndMessage(instance, replicationv1alpha1.Replicating, "volume is marked for resyncing")
+	} else {
+		err = r.updateStatusStateAndMessage(instance, replicationv1alpha1.Replicating, "volume is marked "+string(instance.Spec.ImageState))
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, err
+}
+
+func (r *VolumeReplicationReconciler) updateStatusStateAndMessage(instance *replicationv1alpha1.VolumeReplication, state replicationv1alpha1.ReplicationState, message string) error {
```
sp98, line 211: I would just keeping the method name more generic, like `updateStatus` or `updateReplicationStatus`.  So if want to update more fields in the status in future, then we should not need to change `updateStatusStateAndMessage` name.

controllers/volumereplication_controller.go
```diff
@@ -109,6 +118,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	if secretName != "" && secretNamespace != "" {
 		secret, err = r.getSecret(secretName, secretNamespace)
 		if err != nil {
+			_ = r.updateStatusStateAndMessage(instance, replicationv1alpha1.ReplicationFailure, err.Error())
```
sp98, line 121: why are we ignoring this error?  (and if we are, then we can remove the `_` )

controllers/volumereplication_controller.go
```diff
@@ -84,9 +85,16 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return reconcile.Result{}, err
 	}
 
+	instance.Status.LastStartTime = metav1.Now()
+	if err = r.Client.Update(context.TODO(), instance); err != nil {
```
sp98, line 89: we don't need to add the `LastStartTime` when the cluster is being deleted.

controllers/volumereplication_controller.go
```diff
@@ -84,9 +85,16 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		r.Log.Error(err, "Failed to update status", "Object", instance)
```
sp98, line 90:
```suggestion
		r.Log.Error(err, "Failed to update status", "Object", instance.Name)
```

sp98 (changes requested):

controllers/volumereplication_controller.go
umangachapagain: I was thinking of dumping the entire object, but we could also go with Name and Namespace.

controllers/volumereplication_controller.go
umangachapagain, line 211: You're always ignoring the error returned, so why return it?

controllers/volumereplication_controller.go
```diff
@@ -162,26 +174,49 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	case replicationv1alpha1.Primary:
```
sp98, line 174: How about using the error variable above the switch case and checking it and updating the status once after the switch case. 

```
var replicationErr error 
switch instance.Spec.ImageState {
	case replicationv1alpha1.Primary:
		replicationErr = r.markVolumeAsPrimary(volumeHandle, parameters, secret)
	case replicationv1alpha1.Secondary:
		replicationErr = r.markVolumeAsSecondary(volumeHandle, parameters, secret)
	case replicationv1alpha1.Resync:
		replicationErr := r.resyncVolume(volumeHandle, parameters, secret)
	default:
		replicationErr = fmt.Errorf("unsupported image state %q", instance.Spec.ImageState)
	}


if replicationErr != nil{
   err := r.updateStatusStateAndMessage(instance, replicationv1alpha1.ReplicationFailure, replicationErr.Error())
   if err != nil{
     // log that updating status failed.
     return ctrl.Result{}, err
   }
   return ctrl.Result{}, err
}
```

controllers/volumereplication_controller.go
iamniting: Done

controllers/volumereplication_controller.go
iamniting: I am ignoring the error here because we are returning other err anyway. The other error is much more important than this one.

controllers/volumereplication_controller.go
iamniting: Moved this after deletion logic

controllers/volumereplication_controller.go
iamniting: Done

controllers/volumereplication_controller.go
iamniting: I am not always ignoring it. If you see the last call of the method in reconcile(). I am ignoring err at those places at which we going to return other err anyway.

controllers/volumereplication_controller.go
iamniting: Done

controllers/volumereplication_controller.go
```diff
@@ -87,6 +88,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	// Get VolumeReplicationClass
 	vrcObj, err := r.getVolumeReplicaCLass(instance.Spec.VolumeReplicationClass)
 	if err != nil {
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
```
sp98, line 91: It should be safe to remove `_` above if not using the return value here.

controllers/volumereplication_controller.go
```diff
@@ -158,30 +164,52 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
+	instance.Status.LastStartTime = metav1.Now()
```
sp98, line 167: There can be a case that when you add `LastStartTime`, the resource already has the 'LastCompletionTime` from the previous reconcile. So we will have `LastStartTime` greater than `LasterCompletionTime` in the status.   So we should remove `LastCompletionTime` when we are updating the `LastStartTime`.

controllers/volumereplication_controller.go
iamniting: As far as I know, linter fails in those cases. But let's try

controllers/volumereplication_controller.go
iamniting: Done

controllers/volumereplication_controller.go
iamniting: Screenshot from 2021-03-17 15-15-10

sp98 (approved):

umangachapagain (approved):

PR #45, opened 2021-03-16
enable requeue for resync task when volume is not ready

umangachapagain:
Dropping the idea of arrays of tasks for now as it seems we have a unique requirement for each task.

This PR enables automatic re-syncing of volume until response.Ready is true. 

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	replicationlib "github.com/kube-storage/spec/lib/go/replication"
 	replicationv1alpha1 "github.com/kube-storage/volume-replication-operator/api/v1alpha1"
 	"github.com/kube-storage/volume-replication-operator/controllers/tasks"
 	"github.com/kube-storage/volume-replication-operator/controllers/tasks/replication"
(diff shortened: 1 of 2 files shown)


controllers/volumereplication_controller.go
```diff
@@ -166,16 +167,22 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 
 	case replicationv1alpha1.Secondary:
-		err := r.markVolumeAsSecondary(volumeHandle, parameters, secret)
+		requeueForResync, err := r.markVolumeAsSecondary(types.NamespacedName{Namespace: instance.Namespace, Name: instance.Name}, volumeHandle, parameters, secret)
```
iamniting, line 170: Can not we use `req.NamespacedName` here instead of types.NamespacedName

iamniting (changes requested):

controllers/volumereplication_controller.go
iamniting: Moreover, we can pass the instance itself

umangachapagain: Wouldn't it be better if we fetch the latest instance when we need it?

iamniting: In that case would not it be a problem for example an image state was secondary when the request came while fetching user has changed the image state to the primary. There will be discrepancies as you will have 2 copies of the instance one in your func and the other one in reconcile(). Your func will work according to a new copy of the instance and the reconcile() will work according to the old copy. If we really want to fetch it again we should fetch it into the same pointer IMO.

iamniting (approved):

PR #52, opened 2021-03-22
VR: Fix error "provided secret is empty"

iamniting:
getSecret was returning secret.StringData of type `map[string]string`
which does not have a secret key. Secret key is present in the
secret.Data of type `map[string][]byte`. Which require all funcs to
change the secret type from `map[string]string` to `map[string][]byte`.

This will fix
`{"taskName": "Enable volume replication",
"error": "rpc error: code = Internal desc = provided secret is empty"}`

Depends on https://github.com/kube-storage/spec/pull/17

### DIFF
--- a/controllers/secret.go
+++ b/controllers/secret.go
@@ -25,7 +25,7 @@ import (
 )
 
 // getSecret retrives the secrets based on name and namespace input
-func (r *VolumeReplicationReconciler) getSecret(name, namespace string) (map[string]string, error) {
+func (r *VolumeReplicationReconciler) getSecret(name, namespace string) (map[string][]byte, error) {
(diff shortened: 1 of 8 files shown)


sp98:
both the tests are failing.

Madhu-1:
We don't need a spec change for this one. Get the secrets and convert byte to string and pass it to the driver

Madhu-1:
https://github.com/kubernetes-csi/external-provisioner/blob/bcf1a405b5f6a0fa2aadb80d08c9c2569c4b6912/pkg/controller/controller.go#L1639

iamniting:
Closing it as it will be fixed with #54

PR #54, opened 2021-03-24
VR: Fix error "provided secret is empty"

iamniting:
getSecret was returning secret.StringData of type map[string]string
which does not have a secret key. Secret key is present in the
secret.Data of type map[string][]byte.

Fix the getSecret func via returning secret.Data
and converting it to the map[string]string

### DIFF
--- a/controllers/secret.go
+++ b/controllers/secret.go
@@ -36,5 +36,16 @@ func (r *VolumeReplicationReconciler) getSecret(name, namespace string) (map[str
 		r.Log.Error(err, "error getting secret", "Secret Name", name, "Secret Namespace", namespace)
 		return nil, err
 	}
-	return secret.StringData, nil
+	return convertMap(secret.Data), nil
+}
+
+// convertMap converts map[string][]byte to map[string]string
+func convertMap(oldMap map[string][]byte) map[string]string {
+	newMap := make(map[string]string)
+
+	for key, val := range oldMap {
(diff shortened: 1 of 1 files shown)


sp98 (approved):

umangachapagain (commented):
Why not just return []bytes and then when accessing it, convert it to string?
Wouldn't that be faster?

iamniting:
> Why not just return []bytes and then when accessing it, convert it to string?
> Wouldn't that be faster?

We might get these secrets at other places also. In that case, we have to convert at all places.

umangachapagain:
> > Why not just return []bytes and then when accessing it, convert it to string?
> > Wouldn't that be faster?
> 
> We might get these secrets at other places also. In that case, we have to convert at all places.

Do we use all the key-values in secret? or do we use selected ones?
If former, then this is fine. For latter, it'd save us some extra loops.

iamniting:
> > > Why not just return []bytes and then when accessing it, convert it to string?
> > > Wouldn't that be faster?
> > 
> > 
> > We might get these secrets at other places also. In that case, we have to convert at all places.
> 
> Do we use all the key-values in secret? or do we use selected ones?
> If former, then this is fine. For latter, it'd save us some extra loops.

As of now, we have only two values `user` and `password`.

umangachapagain (approved):

PR #55, opened 2021-03-24
VR: Add more log message in reconcile

iamniting:
As of now reconcile method have very few logs. Which does not really help in case of failures. Add more logs to address the same.

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -154,6 +154,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		if contains(instance.GetFinalizers(), volumeReplicationFinalizer) {
 			err := r.disableVolumeReplication(volumeHandle, parameters, secret)
 			if err != nil {
+				r.Log.Error(err, "Failed to disable replication", "Name", instance.Name)
 				return ctrl.Result{}, err
 			}
 
@@ -193,6 +194,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	if replicationErr != nil {
+		r.Log.Error(replicationErr, "Failed to Replicate", "Name", instance.Name, "ReplicationState", instance.Spec.ReplicationState)
 		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, replicationErr.Error())
 		return ctrl.Result{}, replicationErr
 	}
@@ -201,14 +203,22 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{Requeue: true}, nil
 	}
 
-	instance.Status.LastCompletionTime = getCurrentTime()
+	var msg string
 	if instance.Spec.ReplicationState == replicationv1alpha1.Resync {
-		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked for resyncing")
+		msg = "volume is marked for resyncing"
 	} else {
-		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked "+string(instance.Spec.ReplicationState))
+		msg = fmt.Sprintf("volume is marked %s", string(instance.Spec.ReplicationState))
 	}
 
-	return ctrl.Result{}, err
+	instance.Status.LastCompletionTime = getCurrentTime()
+	err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, msg)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	r.Log.Info(msg, "Name", instance.Name)
+
+	return ctrl.Result{}, nil
 }
 
 func (r *VolumeReplicationReconciler) updateReplicationStatus(instance *replicationv1alpha1.VolumeReplication, state replicationv1alpha1.State, message string) error {
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ bin/*
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+
+# test env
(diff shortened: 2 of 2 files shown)


controllers/volumereplication_controller.go
```diff
@@ -201,11 +203,18 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked for resyncing")
+		msg = "volume is marked for resyncing"
 	} else {
 		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked "+string(instance.Spec.ReplicationState))
+		msg = "volume is marked " + string(instance.Spec.ReplicationState)
+	}
+
+	if err == nil {
+		r.Log.Info(msg, "Name", instance.Name)
 	}
```
sp98, line 219: I would suggest explicitly  returning the error after line 210 and line 213 in case `updateReplicationStatus` fails.  And return `nil` at the end.  Returning `err` at the end might be confusing.

sp98 (changes requested):

controllers/volumereplication_controller.go
```diff
@@ -201,11 +203,18 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	var msg string
 	instance.Status.LastCompletionTime = getCurrentTime()
 	if instance.Spec.ReplicationState == replicationv1alpha1.Resync {
 		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked for resyncing")
+		msg = "volume is marked for resyncing"
```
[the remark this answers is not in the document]

controllers/volumereplication_controller.go
iamniting: Done

controllers/volumereplication_controller.go
```diff
@@ -202,13 +204,26 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+
 	if instance.Spec.ReplicationState == replicationv1alpha1.Resync {
-		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked for resyncing")
+		msg := "volume is marked for resyncing"
+		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, msg)
+		if err != nil {
+			r.Log.Error(err, "Failed to update status", "Name", instance.Name)
+			return ctrl.Result{}, err
+		}
+		r.Log.Info(msg, "Name", instance.Name)
```
sp98, line 215: `msg` can be declared before the if block. And this  log message can be moved after the if block.

controllers/volumereplication_controller.go
```diff
@@ -202,13 +204,26 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	} else {
-		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked "+string(instance.Spec.ReplicationState))
+		msg := "volume is marked " + string(instance.Spec.ReplicationState)
+		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, msg)
+		if err != nil {
+			r.Log.Error(err, "Failed to update status", "Name", instance.Name)
```
sp98, line 220: same error is being logged inside the  `updateReplicationStatus` method.

controllers/volumereplication_controller.go

controllers/volumereplication_controller.go
iamniting: removed

sp98 (approved):

controllers/volumereplication_controller.go
```diff
@@ -201,14 +203,22 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	}
 
-	instance.Status.LastCompletionTime = getCurrentTime()
+	var msg string
 	if instance.Spec.ReplicationState == replicationv1alpha1.Resync {
-		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked for resyncing")
+		msg = "volume is marked for resyncing"
 	} else {
-		err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, "volume is marked "+string(instance.Spec.ReplicationState))
+		msg = "volume is marked " + string(instance.Spec.ReplicationState)
```
Madhu-1, line 210: can we use `fmt.Sprintf`?

controllers/volumereplication_controller.go
```diff
@@ -201,14 +203,22 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	}
 
-	return ctrl.Result{}, err
+	instance.Status.LastCompletionTime = getCurrentTime()
+	err = r.updateReplicationStatus(instance, replicationv1alpha1.Replicating, msg)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	r.Log.Info(msg, "Name", instance.Name)
```
Madhu-1, line 219: This can cause panic?

iamniting: No it should not. Why do you think so

Madhu-1: Looks good, i was thinking something like https://github.com/kube-storage/volume-replication-operator/pull/51 can happen.

controllers/volumereplication_controller.go

Madhu-1 (approved):

controllers/volumereplication_controller.go
iamniting: If you are referring this https://github.com/kube-storage/volume-replication-operator/pull/51 then you can see it was panic when we had 2 args and here we have 3

PR #59, opened 2021-03-26
VR: return err as nil when requeue not required without spec change

iamniting:
If spec.dataSource.kind, spec.replicationState is not correct there is
no need to requeue it again. As recocile will always fails unless user
changes spec.

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -134,8 +134,10 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 		volumeHandle = pv.Spec.CSI.VolumeHandle
 	default:
-		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, ("unsupported datasource kind " + instance.Spec.DataSource.Kind))
-		return ctrl.Result{}, fmt.Errorf("unsupported datasource kind %q", instance.Spec.DataSource.Kind)
+		err = fmt.Errorf("unsupported datasource kind %q", instance.Spec.DataSource.Kind)
+		r.Log.Error(err, "Name", instance.Name)
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
+		return ctrl.Result{}, nil
 	}
 
 	r.Log.Info("volume handle", "Name", volumeHandle)
@@ -191,6 +193,9 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	default:
 		replicationErr = fmt.Errorf("unsupported image state %q", instance.Spec.ReplicationState)
+		r.Log.Error(replicationErr, "unsupported image state", "Name", instance.Name, "ReplicationState", instance.Spec.ReplicationState)
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, replicationErr.Error())
+		return ctrl.Result{}, nil
 	}
 
 	if replicationErr != nil {


controllers/volumereplication_controller.go
```diff
@@ -135,7 +135,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		volumeHandle = pv.Spec.CSI.VolumeHandle
 	default:
 		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, ("unsupported datasource kind " + instance.Spec.DataSource.Kind))
-		return ctrl.Result{}, fmt.Errorf("unsupported datasource kind %q", instance.Spec.DataSource.Kind)
+		return ctrl.Result{Requeue: false}, fmt.Errorf("unsupported datasource kind %q", instance.Spec.DataSource.Kind)
```
sp98, line 138: does this change show the error in the logs for the user?

iamniting: It should do something like 2nd error from `controller-runtime.manager.controller.volumereplication`. But I have not tested 
it out. Anyway, I have made changes.

```
2021-03-26T08:25:28.059Z	ERROR	controllers.VolumeReplication	Failed to Replicate	{"Name": "vr1", "ReplicationState": "Resyncing", "error": "unsupported image state \"Resyncing\""}


2021-03-26T08:25:28.069Z	ERROR	controller-runtime.manager.controller.volumereplication	Reconciler error	{"reconciler group": "replication.storage.openshift.io", "reconciler kind": "VolumeReplication", "name": "vr1", "namespace": "rook-ceph", "error": "unsupported image state \"Resyncing\""}
```

sp98: if not reconciling, then I would suggest just logging the error and returning nil.

controllers/volumereplication_controller.go
```diff
@@ -191,6 +193,9 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	default:
 		replicationErr = fmt.Errorf("unsupported image state %q", instance.Spec.ReplicationState)
+		r.Log.Error(replicationErr, "unsupported image state", "Name", instance.Name, "ReplicationState", instance.Spec.ReplicationState)
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, replicationErr.Error())
+		return ctrl.Result{Requeue: false}, nil
```
sp98, line 198:
```suggestion
		return ctrl.Result{}, nil
```

controllers/volumereplication_controller.go
```diff
@@ -134,8 +134,10 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 		volumeHandle = pv.Spec.CSI.VolumeHandle
 	default:
-		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, ("unsupported datasource kind " + instance.Spec.DataSource.Kind))
-		return ctrl.Result{}, fmt.Errorf("unsupported datasource kind %q", instance.Spec.DataSource.Kind)
+		err = fmt.Errorf("unsupported datasource kind %q", instance.Spec.DataSource.Kind)
+		r.Log.Error(err, "Name", instance.Name)
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
+		return ctrl.Result{Requeue: false}, nil
```
sp98, line 140:
```suggestion
		return ctrl.Result{}, nil
```

sp98 (changes requested):
no need to explicitly mention `Requeue: false` in case we are returning `nil`

Madhu-1 (changes requested):
commit message and PR title  need to be updated

sp98 (approved):

iamniting:
> commit message and PR title need to be updated

Done

Madhu-1 (approved):

PR #60, opened 2021-03-26
VR: Enhance logging of whole operator

iamniting:
As part of this commit did below enhancements
- Add VRName in the logs at all places
- Add other object name at places where ever possible
- Add uniformity in the logs

### DIFF
--- a/controllers/secret.go
+++ b/controllers/secret.go
@@ -31,7 +31,8 @@ func (r *VolumeReplicationReconciler) getSecret(name, namespace string) (map[str
 	err := r.Client.Get(context.TODO(), namespacedName, secret)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			r.Log.Error(err, "Secret not found", "Secret Name", name, "Secret Namespace", namespace)
+			r.Log.Error(err, "secret not found", "Secret Name", name, "Secret Namespace", namespace)
+			return nil, err
 		}
 		r.Log.Error(err, "error getting secret", "Secret Name", name, "Secret Namespace", namespace)
 		return nil, err
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -74,7 +74,9 @@ type VolumeReplicationReconciler struct {
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0/pkg/reconcile
 func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 
-	_ = r.Log.WithValues("controller_volumereplication", req.NamespacedName)
+	prevLogger := r.Log
+	defer func() { r.Log = prevLogger }()
+	r.Log = r.Log.WithValues("Request.Name", req.Name, "Request.Namespace", req.Namespace)
 
 	// Fetch VolumeReplication instance
 	instance := &replicationv1alpha1.VolumeReplication{}
@@ -84,7 +86,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			r.Log.Info("no VolumeReplication resource found")
+			r.Log.Info("volumeReplication resource not found")
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
@@ -105,7 +107,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	err = validatePrefixedParameters(vrcObj.Spec.Parameters)
 	if err != nil {
 		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
-		r.Log.Error(err, "failed to validate prefix parameters", "volumeReplicationClass", instance.Spec.VolumeReplicationClass)
+		r.Log.Error(err, "failed to validate parameters of volumeReplicationClass", "VRCName", instance.Spec.VolumeReplicationClass)
 		return ctrl.Result{}, err
 	}
 	// remove the prefix keys in volume replication class parameters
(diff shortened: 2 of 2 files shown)


controllers/volumereplication_controller.go
```diff
@@ -84,7 +84,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			r.Log.Info("no VolumeReplication resource found")
```
ShyamsundarR, line 87: (question) If you actually used the logger returned from line 77 it would print this information for all logs, without having to add this to each log line. Attempting to learn why that was not chosen.

iamniting: Yeah, Nice question. line 77 did not caught my attention. Maybe we should use that. @umangachapagain @sp98 Can we use that or there was any specific reason to not use it?

Madhu-1:
@umangachapagain PTAL

controllers/volumereplication_controller.go
umangachapagain: We should use that.

controllers/volumereplication_controller.go
```diff
@@ -74,7 +74,9 @@ type VolumeReplicationReconciler struct {
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0/pkg/reconcile
 func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 
-	_ = r.Log.WithValues("controller_volumereplication", req.NamespacedName)
+	prevLogger := r.Log
```
Madhu-1, line 77: why these changes are required?

controllers/volumereplication_controller.go
iamniting: Done

controllers/volumereplication_controller.go
iamniting: because of this reason https://github.com/openshift/ocs-operator/pull/990#discussion_r558330313

Moreover `VolumeReplicationReconciler` gets generated once when we start the operator and we are changing its value r.log field which will be retained even when it returns from reconciling. So it does not make sense to keep the requested value in r.log IMO.

controllers/secret.go
```diff
@@ -32,6 +32,7 @@ func (r *VolumeReplicationReconciler) getSecret(name, namespace string) (map[str
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			r.Log.Error(err, "Secret not found", "Secret Name", name, "Secret Namespace", namespace)
```
sp98, line 34:
```suggestion
			r.Log.Error(err, "secret not found", "Secret Name", name, "Secret Namespace", namespace)
```

lets keep every message in the same case (lower case) for consistency.

controllers/volumereplication_controller.go
```diff
@@ -130,58 +132,58 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		_, pv, err := r.getPVCDataSource(nameSpacedName)
 		if err != nil {
 			_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
-			r.Log.Error(err, "failed to get dataSource for PVC", "dataSourceName", nameSpacedName.Name)
+			r.Log.Error(err, "failed to get PVC", "PVCName", instance.Spec.DataSource.Name)
```
sp98, line 135: do you want to add instance name here as well?

controllers/volumereplication_controller.go
```diff
@@ -130,58 +132,58 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		}
-		r.Log.Info("volumeReplication object is terminated, skipping reconciliation", "objectName", instance.Name)
+		r.Log.Info("volumeReplication object is terminated, skipping reconciliation")
 		return ctrl.Result{}, nil
 	}
 
 	instance.Status.LastStartTime = getCurrentTime()
 	if err = r.Client.Update(context.TODO(), instance); err != nil {
-		r.Log.Error(err, "Failed to update status", "Object", instance.Name)
+		r.Log.Error(err, "Failed to update status")
```
sp98, line 180:
```suggestion
		r.Log.Error(err, "failed to update status")
```

controllers/volumereplication_controller.go
```diff
@@ -130,58 +132,58 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		return reconcile.Result{}, err
 	}
 
 	// enable replication on every reconcile
 	if err = r.enableReplication(volumeHandle, parameters, secret); err != nil {
-		r.Log.Error(err, "Failed to enable replication", "Object", instance.Name)
+		r.Log.Error(err, "Failed to enable replication")
```
sp98, line 186:
```suggestion
		r.Log.Error(err, "failed to enable replication")
```

Madhu-1 (approved):

controllers/volumereplication_controller.go
```diff
@@ -238,7 +240,7 @@ func (r *VolumeReplicationReconciler) updateReplicationStatus(instance *replicat
 	instance.Status.State = state
 	instance.Status.Message = message
 	if err := r.Client.Status().Update(context.TODO(), instance); err != nil {
-		r.Log.Error(err, "Failed to update status", "Object", instance.Name)
+		r.Log.Error(err, "Failed to update status")
```
sp98, line 243:
```suggestion
		r.Log.Error(err, "failed to update status")
```

sp98 (changes requested):
nit

controllers/volumereplication_controller.go
iamniting: Nope, now we have a instance name by default

sp98 (changes requested):
one last thing . Log in the `requeueForResync` if block   [here](https://github.com/csi-addons/volume-replication-operator/blob/a918564f6b70be59ba0600944ced11f81402d546/controllers/volumereplication_controller.go#L218) would be very useful.

iamniting:
> one last thing . Log in the `requeueForResync` if block [here](https://github.com/csi-addons/volume-replication-operator/blob/a918564f6b70be59ba0600944ced11f81402d546/controllers/volumereplication_controller.go#L218) would be very useful.

done, updated status also

sp98 (approved):

controllers/volumereplication_controller.go
```diff
@@ -200,19 +202,21 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		requeueForResync, replicationErr = r.resyncVolume(instance, volumeHandle, parameters, secret)
 
 	default:
-		replicationErr = fmt.Errorf("unsupported image state %q", instance.Spec.ReplicationState)
-		r.Log.Error(replicationErr, "unsupported image state", "Name", instance.Name, "ReplicationState", instance.Spec.ReplicationState)
+		replicationErr = fmt.Errorf("unsupported image state")
+		r.Log.Error(replicationErr, "given image state is not supported", "ReplicationState", instance.Spec.ReplicationState)
```
Madhu-1, line 206: the image needs to be changed to volume.

PR #61, opened 2021-03-29
add conditions in CR status

sp98:
Conditions in the Custome Resource Status helps to provide details about a particular volume replication state (Primary, Secondary or Resync)

hackmd design - https://hackmd.io/XbtxPr1GQ7CmTGW1VVFSAg


controllers/status.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	ConditionCompleted = "Completed"
+	ConditionDegraded  = "Degraded"
+	ConditionResyncing = "Resyncing"
```
iamniting, line 27: I think this should be resncTrigerred.

controllers/volumereplication_controller.go
```diff
@@ -227,8 +227,8 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	return ctrl.Result{}, nil
 }
 
-func (r *VolumeReplicationReconciler) updateReplicationStatus(instance *replicationv1alpha1.VolumeReplication, state replicationv1alpha1.State, message string) error {
-	instance.Status.State = state
+func (r *VolumeReplicationReconciler) updateReplicationStatus(instance *replicationv1alpha1.VolumeReplication, message string) error {
```
iamniting, line 230: It will update the state when this will get called for the first time in reconcile. This should not be the case IMO. We should update the state only when we are successfully able to achieve/change the state.

controllers/status.go
```diff
@@ -0,0 +1,134 @@
...
+	Demoted         = "Demoted"
+	FailedToPromote = "FailedToPromote"
+	FailedToDemote  = "FailedToDemote"
+	Error           = "Error"
+	Healthy         = "Healthy"
+	ResycTriggered  = "ResyncTriggered"
+	FailedToResync  = "FailedToResync"
+)
+
+func setPromotedCondition(conditions *[]conditionsv1.Condition) {
```
iamniting, line 42: Will this show all three conditions in the status? If not we have set 3rd condition also.

iamniting (changes requested):

sp98:
PR not ready for review yet.

controllers/status.go
sp98: `Resyncing` is the conditions type.  We have `ResyncTriggered`, `FailedToResync` and `NotResyncing` as the reason.

controllers/volumereplication_controller.go
sp98: its ok to update the `state` when we reconcile for the first time. `Status.State` represents what's the last operation user performed on the CR. The conditions will reflect whether we are successfully able to achieve/change the state.

controllers/status.go
sp98: I've updated the conditions to show all the three conditions in this case as well.

controllers/status.go
```diff
@@ -0,0 +1,155 @@
...
+	FailedToPromote = "FailedToPromote"
+	FailedToDemote  = "FailedToDemote"
+	Error           = "Error"
+	Healthy         = "Healthy"
+	ResycTriggered  = "ResyncTriggered"
+	FailedToResync  = "FailedToResync"
+	NotResyncing    = "NotResyncing"
+)
+
+// sets conditions when image was promoted successfully
```
ShyamsundarR, line 43: Request that we consider setting the `observedGeneration` as well when setting the conditions, to ensure we can correlate with the latest generation of the CR.

ShyamsundarR, line 43: Also, we should consider adding timestamps as well. For example a `Completed` condition can have happened earlier, and now the resource is `Degraded` this can be gleaned from status time stamps on when each condition was observed.

controllers/status.go
```diff
@@ -0,0 +1,155 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
```
Madhu-1, line 20: is this not available in the kubernetes package?

controllers/volumereplication_controller.go
```diff
@@ -466,6 +472,16 @@ func (r *VolumeReplicationReconciler) hasKnownGRPCError(tasks []*tasks.TaskSpec,
 	return false
 }
 
+func getReplicationState(instance *replicationv1alpha1.VolumeReplication) replicationv1alpha1.State {
+	var replicationState replicationv1alpha1.State
+	if instance.Spec.ReplicationState == replicationv1alpha1.Primary {
```
Madhu-1, line 477: convert this to a switch case?

sp98: done

controllers/status.go
sp98: using from K8s package now.

controllers/status.go
sp98: added observedGenerations with conditions.

api/v1alpha1/volumereplication_types.go
```diff
@@ -39,14 +39,14 @@ const (
...
+	PrimaryState State = "Primary"
 
-	// ReplicationFailure means the last replication operation failed
-	ReplicationFailure State = "Failed"
+	// SecondaryState represents the Secondary replication state
+	SecondaryState State = "Secondary"
 
-	//Resyncing means that the volume is resyncing
-	Resyncing State = "Resyncing"
+	// UnknownState represents the Secondary replication state
```
iamniting, line 48: It should be 

// UnknownState represents the Uknown replication state

controllers/status.go
```diff
@@ -0,0 +1,196 @@
...
+		Type:               ConditionResyncing,
+		Reason:             NotResyncing,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionFalse,
+	})
+
+}
+
+// sets conditions when image resync was triggered successfully
+func setResyncCondition(conditions *[]metav1.Condition, observedGeneration int64) {
```
iamniting, line 135: should not we set all 3 conditions here?

controllers/volumereplication_controller.go
```diff
@@ -486,6 +472,19 @@ func (r *VolumeReplicationReconciler) hasKnownGRPCError(tasks []*tasks.TaskSpec,
 	return false
 }
 
+func getReplicationState(instance *replicationv1alpha1.VolumeReplication) replicationv1alpha1.State {
```
iamniting, line 475: This will give you the requested state in the spec and won't give the actual state of the image. We should get the actual state of the image somehow and show that in the status part not the one requested.

The other way around we can update the state exact after promote/demote success that way we can show the actual state of the image.

iamniting (approved):

api/v1alpha1/volumereplication_types.go
sp98: fixed.

controllers/status.go
sp98: added all three conditions here.

controllers/volumereplication_controller.go
sp98: That's a good point.  Since not getting the actual state from the image, made below changes: 
For successful reconcile, updating the `status.state` to use `spec.ReplicationState`.  For failure scenarios, using the current `status.State` only (or `unkown` in case `status.state` is empty, say, for first reconcile).

controllers/status.go
sp98: `Status.Conditions[Completed][Timestamp]` is updated whenever there is change in `Status.Conditions[Completed][Status]`

iamniting (approved):

controllers/status.go
```diff
@@ -0,0 +1,219 @@
...
+	setStatusCondition(conditions, metav1.Condition{
+		Type:               ConditionResyncing,
+		Reason:             NotResyncing,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionFalse,
+	})
+
+}
+
+// sets conditions when image resync was triggered successfully
```
Madhu-1, line 135: replace image with volume in comments.

Madhu-1:
@umangachapagain PTAL

umangachapagain (approved):

PR #63, opened 2021-03-31
VR: Fix markVolumeAsPrimary to return err in case of !isKnownError

iamniting:
markVolumeAsPrimary was not returning an error when we have an error
which is not known to us. Because of which we were getting nil in
recocile and updating the status of the VR to primary.


controllers/volumereplication_controller.go
```diff
@@ -304,6 +304,16 @@ func (r *VolumeReplicationReconciler) markVolumeAsPrimary(volumeReplicationObjec
 	resp := tasks.RunAll(promoteVolumeTasks)
 
 	isKnownError := r.hasKnownGRPCError(promoteVolumeTasks, resp)
+
+	if !isKnownError {
```
Madhu-1, line 308: Add this in else part for existing if isknownerror

iamniting: Done

Madhu-1 (approved):

PR #64, opened 2021-04-01
enable volume replication on every reconcile

sp98:
Replication was only enabled with Primary Replicaiton State. CR could be created with other states as well, like `Secondary`. This PR enables replication on the every  reconcile.

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -179,6 +179,13 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return reconcile.Result{}, err
 	}
 
+	// enable replication on every reconcile
+	if err = r.enableReplication(volumeHandle, parameters, secret); err != nil {
+		r.Log.Error(err, "Failed to enable replication", "Object", instance.Name)
+		_ = r.updateReplicationStatus(instance, replicationv1alpha1.ReplicationFailure, err.Error())
+		return reconcile.Result{}, err
+	}
+
 	var replicationErr error
 	var requeueForResync bool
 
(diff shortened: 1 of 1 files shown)


controllers/volumereplication_controller.go
```diff
@@ -179,6 +179,15 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return reconcile.Result{}, err
 	}
 
+	// empty `Status.State` conveys that this is the first reconcile, so we enable replication.
+	if instance.Status.State == "" {
+		if err = r.enableReplication(volumeHandle, parameters, secret); err != nil {
```
iamniting, line 184: If there is no error let's update the state here. So that it won't go in this if the condition again

iamniting (changes requested):

controllers/volumereplication_controller.go
sp98: planning to do that in the status conditions PR - https://github.com/kube-storage/volume-replication-operator/pull/61

sp98: I'll be updating the conditions and state in the https://github.com/kube-storage/volume-replication-operator/pull/61 PR as well.

sp98: There is another issue that here that if the `enabling` replication fails, it won't be called again because we are changing the `replicationv1alpha1.State` to `ReplicationFailure`. But that will be fixed in the other PR where we are changing the replication states completely.

umangachapagain (approved):

controllers/volumereplication_controller.go
```diff
@@ -179,6 +179,15 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return reconcile.Result{}, err
 	}
 
+	// empty `Status.State` conveys that this is the first reconcile, so we enable replication.
+	if instance.Status.State == "" {
```
Madhu-1, line 183: lets not depend on the state. Even if the state is secondary, send a request for enabling,Demote and Resync. as spec need to be idempotent  it should return success

iamniting (approved):

controllers/volumereplication_controller.go
sp98: so enable replication on very reconcile?

Madhu-1: if we can depend on the status This is fine, if we don't want to depend on the status we won't be able to reconcile if someone disabled mirroring in backend or something changed in the backend.

sp98: ok. makes sense to not depend on status.state.  Removed it.

Madhu-1 (approved):

controllers/volumereplication_controller.go
sp98: This should no longer be a problem as we are enabling replication on every reconcile. See [comment](https://github.com/kube-storage/volume-replication-operator/pull/64#discussion_r606921994)

umangachapagain (approved):

PR #76, opened 2021-04-27
requeue the request if the volume is marked as secondary

Madhu-1:
when there is a request to mark the volume as secondary, First demote the image and then requeue the request with a delay of `15 seconds`. This delay is introduced because some storage system needs some time to identify the volume needs resync
or not, when the Reconcile happens again it will demote the volume and then sends a request to check the volume needs resync or not.

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -196,14 +196,39 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		replicationErr = r.markVolumeAsPrimary(instance, logger, volumeHandle, parameters, secret)
 
 	case replicationv1alpha1.Secondary:
-		replicationErr = r.markVolumeAsSecondary(instance, logger, volumeHandle, parameters, secret)
-		// resync volume if successfully marked Secondary
-		if replicationErr == nil {
-			err := r.updateReplicationStatus(instance, logger, getReplicationState(instance), "volume is marked secondary")
-			if err != nil {
-				return ctrl.Result{}, err
+		// For the first time, mark the volume as secondary and requeue the
+		// request. For some storage providers it takes some time to determine
+		// whether the volume need correction example:- correcting split brain.
+		if instance.Status.State != replicationv1alpha1.SecondaryState {
+			replicationErr = r.markVolumeAsSecondary(instance, logger, volumeHandle, parameters, secret)
+			if replicationErr == nil {
+				err := r.updateReplicationStatus(instance, logger, getReplicationState(instance), "volume is marked secondary")
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				logger.Info("volume is not ready to use")
+				setOnlyDegradedCondition(&instance.Status.Conditions, instance.Generation)
+				err = r.updateReplicationStatus(instance, logger, getCurrentReplicationState(instance), "volume is degraded")
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				return ctrl.Result{
+					Requeue: true,
+					// Setting Requeue time for 15 seconds
+					RequeueAfter: time.Duration(time.Second * 15),
+				}, nil
 			}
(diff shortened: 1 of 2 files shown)


controllers/volumereplication_controller.go
```diff
@@ -196,14 +196,39 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+					return ctrl.Result{}, err
+				}
+				return ctrl.Result{
+					Requeue: true,
+					// Setting Requeue time for 15 seconds
+					RequeueAfter: time.Duration(time.Second * 15),
+				}, nil
+			}
+		} else {
+			replicationErr = r.markVolumeAsSecondary(instance, logger, volumeHandle, parameters, secret)
```
iamniting, line 222: I think we don't need to call this method again

controllers/volumereplication_controller.go
```diff
@@ -196,14 +196,39 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-		replicationErr = r.markVolumeAsSecondary(instance, logger, volumeHandle, parameters, secret)
-		// resync volume if successfully marked Secondary
-		if replicationErr == nil {
-			err := r.updateReplicationStatus(instance, logger, getReplicationState(instance), "volume is marked secondary")
-			if err != nil {
-				return ctrl.Result{}, err
+		// For the first time, Mark the volume as secondary and requeue the
+		// request, For some storage providers it takes some time to determine
+		// whether the volume need correction example:- correcting split brain.
+		if instance.Status.State == "" || instance.Status.State != replicationv1alpha1.SecondaryState {
```
iamniting, line 202: is not only second condition sufficient here

iamniting (changes requested):

controllers/volumereplication_controller.go
Madhu-1: This is to make  sure that we are marking the image as secondary if something as changed in backend

controllers/volumereplication_controller.go
Madhu-1: No, this is to handle the VR creation with secondary replicationState

iamniting: still, 2nd condition should be enough IMO `instance.Status.State != replicationv1alpha1.SecondaryState`

If you have a new CR with the `""` state and only the above condition it will go inside condition as the state is not equal to secodary.

iamniting (approved):

controllers/volumereplication_controller.go
Madhu-1: Yes looks like the second condition  is enough, fixed it now

controllers/volumereplication_controller.go
```diff
@@ -196,14 +196,39 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-		replicationErr = r.markVolumeAsSecondary(instance, logger, volumeHandle, parameters, secret)
-		// resync volume if successfully marked Secondary
-		if replicationErr == nil {
-			err := r.updateReplicationStatus(instance, logger, getReplicationState(instance), "volume is marked secondary")
-			if err != nil {
-				return ctrl.Result{}, err
+		// For the first time, Mark the volume as secondary and requeue the
+		// request, For some storage providers it takes some time to determine
+		// whether the volume need correction example:- correcting split brain.
+		if instance.Status.State != replicationv1alpha1.SecondaryState {
```
sp98, line 202: `instance.Status.State` would be `nil` for the first reconcile. And  if the user creates the volume replication resource with replication state as `Secondary`, this if condition might fail. So IMO, we should check `instance.Status` not nil as well. What do you think?

sp98 (changes requested):

controllers/volumereplication_controller.go
Madhu-1: `instance.Status` is not a pointer I think we don't need nil check here, i didn't saw any error when CR is created with secondary state

sp98 (approved):

Issue #91, opened 2021-07-08
Disable mirroring on delete, only if the VR states its stat is Primary


ShyamsundarR:
The [code](https://github.com/csi-addons/volume-replication-operator/blob/7549c3c7d1414bb9b74c3af7e4b61443ee4d6285/controllers/volumereplication_controller.go#L159) invokes disabling replication when a CR is deleted.

If a CR is Secondary, then the assumption is it is Primary somewhere else, or at least there is no conclusive action to actually disable mirroring.

The request is to disable mirroring when a VR is deleted as Primary and to just delete the CR (without disabling mirroring) when the state is Secondary.

PR #94, opened 2021-07-09
replication: disable replication if its primary state

Madhu-1:
if a CR is Secondary, then the assumption is it is Primary somewhere else, or at least there is no conclusive action to actually disable mirroring.

Allowing the Deletion of CR if the image is in a secondary state. and the Replication will be disabled when the VR is deleted on the PVC is in a primary state.

fixes: #91


sp98 (approved):

ShyamsundarR (commented):
LGTM! I will reserve approve votes from others, hence not adding the +1

Madhu-1:
@iamniting PTAL

iamniting (approved):

PR #96, opened 2021-07-14
ci: push images only on volume-replication-operator repo

Madhu-1:
this commit avoids running push and publish jobs from the forked repo.

### DIFF
--- a/.github/workflows/docker-publish.yaml
+++ b/.github/workflows/docker-publish.yaml
@@ -8,6 +8,7 @@ jobs:
   push:
     name: Push Image
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


iamniting (approved):
Even though jobs are running in forked repos. I am expecting that they will fail eventually. Because those GitHub accounts do not have quay access to push those images right?

Madhu-1:
> Even though jobs are running in forked repos. I am expecting that they will fail eventually. Because those GitHub accounts do not have quay access to push those images right?

Yes correct

Madhu-1:
@mergifyio refresh

PR #99, opened 2021-08-05
Disable stack trace for error and above error level logs

raghavendra-talur:
One can still get stack trace for Warn and Error levels by running the operator in development mode by setting the env VOL_REP_DEVELOPER true.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/onsi/ginkgo v1.14.1
 	github.com/onsi/gomega v1.10.2
 	github.com/stretchr/testify v1.5.1
+	go.uber.org/zap v1.15.0
 	google.golang.org/grpc v1.32.0
 	k8s.io/api v0.19.2
 	k8s.io/apimachinery v0.19.2
--- a/main.go
+++ b/main.go
@@ -20,12 +20,14 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	uberzap "go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -62,6 +64,7 @@ func main() {
 	var leaderElectionNamespace string
 	var enableLeaderElection bool
 	var probeAddr string
+	var opts zap.Options
 
 	if strings.EqualFold(os.Getenv("DEVELOPMENT_MODE"), "true") {
 		developmentMode = true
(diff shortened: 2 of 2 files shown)


main.go
```diff
@@ -61,6 +64,11 @@ func main() {
 	var leaderElectionNamespace string
 	var enableLeaderElection bool
 	var probeAddr string
+	var opts zap.Options
+
+	if strings.EqualFold(os.Getenv("VOL_REP_DEVELOPER"), "true") {
```
Madhu-1, line 69: `VOL_REP_DEVELOPER` to `DEVELOPMENT_MODE` or `RUN_AS_DEVELOPER`?

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/onsi/ginkgo v1.14.1
 	github.com/onsi/gomega v1.10.2
 	github.com/stretchr/testify v1.5.1
+	go.uber.org/zap v1.15.0
```
Madhu-1, line 12: can  we use latest version ie `v1.18.1`

Madhu-1 (commented):
small suggestion LGTM

main.go
```diff
@@ -73,9 +81,22 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	opts := zap.Options{
-		Development: true,
+
+	opts = zap.Options{
+		ZapOpts: []uberzap.Option{
+			uberzap.AddCaller(),
```
sp98, line 87: I'm trying to understand the need for uberzap here.  The [operator-sdk ](https://sdk.operatorframework.io/docs/building-operators/golang/references/logging/#default-zap-logger)suggests using `--zap-devel` which have the defaults that we need.

go.mod
raghavendra-talur: I chose v1.15.0 as that is the version that the logger framework in controller-runtime is pulling. Matching it will keep surprises to minimum. We only need the import to get the Options, functionality is still handled by the controller runtime.

main.go
raghavendra-talur: It is usually recommended that you prefix the ENV with something specific to your binary/project. The recommendation is not as applicable to processes that mainly run in the containers where the processes are isolated in comparison to the processes that are generally run alongside many other processes.

Happy to do the change to DEVELOPMENT_MODE though.

main.go
raghavendra-talur: The `--zap-devel` flag will enable the stacktrace for all levels WARN and above. We don't want that; I couldn't find a Zap flag that equates to {show caller in the logs, don't show stack trace unless it is panic level}, that is the reason I had to craft a ZapOpts object.

sp98 (approved):

main.go
```diff
@@ -73,9 +81,22 @@ func main() {
...
+			uberzap.AddStacktrace(uberzap.PanicLevel),
+		},
 	}
+
+	if developmentMode {
```
iamniting, line 92: Can we use `else` for the above declaration of `opts` here?

iamniting (changes requested):

main.go
raghavendra-talur: I would prefer not to do that.

Currently the code reads like this 
```
set defaults
if dev_mode {
   override defaults
}
 ```
 
 Based on your suggestion, we can do either 
 ```
 if dev_mode {
    dev_mode settings
} else {
   default settings
}
```

OR 

 ```
 if !dev_mode {
    default settings
} else {
   dev_mode settings
}
```

In both the cases, the defaults get hidden inside the if/else block and I would prefer to have them outside. Let me know what you think..

iamniting: makes sense

iamniting (dismissed):

main.go
raghavendra-talur: Done

iamniting (approved):

Madhu-1 (approved):

PR #103, opened 2021-08-11
Revert "replication: disable replication if its primary state"

Madhu-1:
This reverts #94. More discussions on https://github.com/csi-addons/volume-replication-operator/issues/100

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -167,15 +167,12 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 	} else {
 		if contains(instance.GetFinalizers(), volumeReplicationFinalizer) {
-			// Disable replication only if it is in primary state and PV object
-			// is present.
-			if instance.Spec.ReplicationState == replicationv1alpha1.Primary && pvErr == nil {
-				err := r.disableVolumeReplication(logger, volumeHandle, parameters, secret)
-				if err != nil {
-					logger.Error(err, "failed to disable replication")
-					return ctrl.Result{}, err
-				}
+			err := r.disableVolumeReplication(logger, volumeHandle, parameters, secret)
+			if err != nil {
+				logger.Error(err, "failed to disable replication")
+				return ctrl.Result{}, err
 			}
+
 			logger.Info("removing finalizer from volumeReplication object", "Finalizer", volumeReplicationFinalizer)
 			// once all finalizers have been removed, the object will be deleted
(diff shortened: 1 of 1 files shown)


ShyamsundarR (approved):

ShyamsundarR:
Would like to see the ceph-csi changes to handle disable as secondary, so that the workflow is not disturbed, but that may come later I guess.

controllers/volumereplication_controller.go
```diff
@@ -167,15 +167,12 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		if contains(instance.GetFinalizers(), volumeReplicationFinalizer) {
-			// Disable replication only if it is in primary state and PV object
-			// is present.
-			if instance.Spec.ReplicationState == replicationv1alpha1.Primary && pvErr == nil {
-				err := r.disableVolumeReplication(logger, volumeHandle, parameters, secret)
-				if err != nil {
-					logger.Error(err, "failed to disable replication")
-					return ctrl.Result{}, err
-				}
+			err := r.disableVolumeReplication(logger, volumeHandle, parameters, secret)
```
ShyamsundarR, line 170: We need to call disable without checking state of the VR in the case of Delete as well, which currently calls disable only if it is Primary, right?

Madhu-1: Yes correct I will fix it as part of https://github.com/csi-addons/volume-replication-operator/pull/104

Madhu-1: sorry, discard my last comment. This is reverting back the change and there is no check on the VR state to call the disableReplication. Am I missing something?

Madhu-1:
> Would like to see the ceph-csi changes to handle disable as secondary, so that the workflow is not disturbed, but that may come later I guess.

Yes working on the cephcsi changes will send PR soon for that one.

controllers/volumereplication_controller.go
ShyamsundarR: We should call disable unconditionally [here](https://github.com/csi-addons/volume-replication-operator/blob/00f42351bccb8020ef90d36e65d4320959154285/controllers/volumereplication_controller.go#L172-L173) as well, right? As the storage provider needs to handle if it is `Primary` and do the appropriate action.

Madhu-1: isn't it doing at the same place?

iamniting (approved):

controllers/volumereplication_controller.go
ShyamsundarR: yes, ack! sorry wires crossed between PR reviews.

Issue #97, opened 2021-08-04
missing CRDs cause crash on startup instead of failure on replication creation


oriyarde:
the list of sidecars of our CSI controller deployment is constant, i.e. doesn't change according to existence of CRDs on the cluster.

so we would like to be able to deploy the volume replication operator even when the [replication CRDs](https://github.com/csi-addons/volume-replication-operator/tree/v0.1.0/config/crd/bases) don't exist on the cluster.
this is the error we get on startup:
```
2021-08-04T12:22:20.018Z        ERROR   controller-runtime.source       if kind is a CRD, it should be installed before calling Start   {"kind": "VolumeReplication.replication.storage.openshift.io", "error": "no matches for kind \"VolumeReplication\" in version \"replication.storage.openshift.io/v1alpha1\""}
github.com/go-logr/zapr.(*zapLogger).Error
        /go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132
sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/source/source.go:117
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/internal/controller/controller.go:159
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/internal/controller/controller.go:205
sigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).startRunnable.func1
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/manager/internal.go:691
2021-08-04T12:22:20.019Z        ERROR   setup   problem running manager {"error": "no matches for kind \"VolumeReplication\" in version \"replication.storage.openshift.io/v1alpha1\""}
github.com/go-logr/zapr.(*zapLogger).Error
        /go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132
main.main
        /workspace/main.go:134
runtime.main
        /usr/local/go/src/runtime/proc.go:204
```
[...]

PR #108, opened 2021-08-29
wait for volume replication crds

matancarmeli7:
This PR will allow the volume replication operator to start and not fail when `volumereplicationclasses` and `volumereplications` crds are not in the cluster.
It will wait until they will create.
This PR is related to #97 issue.

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -25,8 +25,11 @@ import (
 	"google.golang.org/grpc/codes"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -42,6 +45,8 @@ import (
 
 const (
 	pvcDataSource = "PersistentVolumeClaim"
+	volumeReplicationClass = "VolumeReplicationClass"
+	volumeReplication = "VolumeReplication"
 )
 
 var (
@@ -297,6 +302,12 @@ func (r *VolumeReplicationReconciler) updateReplicationStatus(instance *replicat
 // SetupWithManager sets up the controller with the Manager.
 func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.DriverConfig) error {
 
+	err := r.waitForCrds()
+	if err != nil {
+		r.Log.Error(err, "failed to wait for crds")
+		return err
+	}
+
 	pred := predicate.GenerationChangedPredicate{}
 
 	r.DriverConfig = cfg
@@ -315,6 +326,48 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&replicationv1alpha1.VolumeReplication{}).
 		WithEventFilter(pred).Complete(r)
+}
+
+func (r *VolumeReplicationReconciler) waitForCrds() error {
+	logger := r.Log.WithName("checkingDependencies")
+
+	err := r.waitForVolumeReplicationResource(logger, volumeReplicationClass)
+	if err != nil {
+		logger.Error(err, "failed to wait for VolumeReplicationClass CRD")
+		return err
+	}
+
+	err = r.waitForVolumeReplicationResource(logger, volumeReplication)
+	if err != nil {
+		logger.Error(err, "failed to wait for VolumeReplication CRD")
+		return err
+	}
+	return nil
+}
+
+func (r *VolumeReplicationReconciler) waitForVolumeReplicationResource(logger logr.Logger, resourceName string) error {
+	isResourceExists := false
+	unstructuredResource := &unstructured.UnstructuredList{}
+	unstructuredResource.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   replicationv1alpha1.GroupVersion.Group,
+		Kind:    resourceName,
+		Version: replicationv1alpha1.GroupVersion.Version,
+	})
+	for !isResourceExists {
+		err := r.Client.List(context.TODO(), unstructuredResource)
+		if err != nil {
+			if meta.IsNoMatchError(err){
+				logger.Info("resource does not exist", "Resource",  resourceName)
+				time.Sleep(5 * time.Second)
+			} else {
+				logger.Error(err, "got an unexpected error while waiting for resource", "Resource", resourceName)
+				return err
+			}
+		} else {
+			isResourceExists = true
+		}
+	}
+	return nil
 }
 
 // markVolumeAsPrimary defines and runs a set of tasks required to mark a volume as primary


controllers/volumereplication_controller.go
```diff
@@ -317,6 +325,64 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+
+func (r *VolumeReplicationReconciler) waitForVolumeReplicationClass(logger logr.Logger) error {
+
+	vrcObj := &replicationv1alpha1.VolumeReplicationClass{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: ""}, vrcObj)
+	if err != nil {
+		if meta.IsNoMatchError(err){
+			logger.Info("Resource", "VolumeReplicationClass", "does not exist")
+			time.Sleep(5 * time.Second)
+			err = r.waitForVolumeReplicationClass(logger)
```
ShyamsundarR, line 354: Instead of recursion we can just use a for(ever) loop here and break on errors or if the CRD is found?

I do not know ho recursion in go works, but this will accumulate a function on the stack every 5 seconds, and for a sidecar that expects a CRD to come at some point in time in the future, this maybe quite a lot of accumulations.

(applies to `waitForVolumeReplication` as well)

controllers/volumereplication_controller.go
```diff
@@ -317,6 +325,64 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+		logger.Error(err, "Failed to watch VolumeReplication")
+		return err
+	}
+	return nil
+}
+
+func (r *VolumeReplicationReconciler) waitForVolumeReplicationClass(logger logr.Logger) error {
+
+	vrcObj := &replicationv1alpha1.VolumeReplicationClass{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: ""}, vrcObj)
```
ShyamsundarR, line 349: Instead of Get (which is for a specific resource) should we just attempt a List here? (@Madhu-1  thoughts?)

It seems odd that we expect errors that states the object is invalid (or unknown), than Name is empty initially, which a List can help disambiguate.

(applies to `waitForVolumeReplication` as well)

matancarmeli7: done

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
```diff
@@ -317,6 +325,70 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+		if err != nil {
+			if meta.IsNoMatchError(err){
+				logger.Info("Resource", "VolumeReplication", "does not exist")
+				time.Sleep(5 * time.Second)
+			} else if ! strings.Contains(err.Error(), "cache is not started"){
+				logger.Error(err, "Got an unexpected error while waiting for", "Resource", "VolumeReplication")
+				return err
+			} else {
+				isVolumeReplicationCrdExists = true
+			}
```
Madhu-1, line 384: why this else block is required?

controllers/volumereplication_controller.go
```diff
@@ -317,6 +325,70 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+
+	isVolumeReplicationCrdExists := false
+	for ! isVolumeReplicationCrdExists {
+		vrOBJ := &replicationv1alpha1.VolumeReplicationList{}
+		err := r.Client.List(context.TODO(), vrOBJ)
+		if err != nil {
+			if meta.IsNoMatchError(err){
+				logger.Info("Resource", "VolumeReplication", "does not exist")
+				time.Sleep(5 * time.Second)
+			} else if ! strings.Contains(err.Error(), "cache is not started"){
```
Madhu-1, line 379: can you please add a comment here?
will this happen on all SDK versions or specific SDK versions?
are we considering `cache is not started` as the success message here and why?

Madhu-1 (changes requested):

controllers/volumereplication_controller.go
Madhu-1: okay this is to consider `cache not started` as successfull

controllers/volumereplication_controller.go
matancarmeli7: > are we considering `cache is not started` as the success message here and why?
>
if we get `cache is not started` error, that means we have the wanted crd in the cluster

matancarmeli7: > can you please add a comment here?
>
what will be the comment on? why we put the `cache is not started` there?

Madhu-1: if we get `cache is not started` it doesn't mean the CRD is present rather it's a different error? can we just consider this as a normal error message to let the operator restart? or use a different client as mentioned at https://github.com/openshift-knative/serverless-operator/pull/658 https://github.com/jaegertracing/jaeger-operator/issues/809#issue-532751170

matancarmeli7: I tried what you suggested but it didn't work well.
I used [unstructured](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured) module and it solves the `cache is not started` problem

controllers/volumereplication_controller.go
```diff
@@ -42,6 +45,8 @@ import (
 
 const (
 	pvcDataSource = "PersistentVolumeClaim"
+	VolumeReplicationClass = "VolumeReplicationClass"
+	VolumeReplication = "VolumeReplication"
 )
```
Madhu-1, line 50: variables should be unexported should start with smallcase

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+		return err
+	}
+	return nil
+}
+
+func (r *VolumeReplicationReconciler) waitForVolumeReplicationResource(logger logr.Logger, resourceName string) error {
+
+	isVolumeReplicationResourceExists := false
+	for ! isVolumeReplicationResourceExists {
+		unstructuredResource := &unstructured.UnstructuredList{}
```
Madhu-1, line 353: this need to be initialized once move this out of for loop.

Madhu-1 (commented):
this looks much better.

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
matancarmeli7: done

Madhu-1 (approved):

Madhu-1:
@matancarmeli7 can you please squash all commits into one?

Madhu-1 (changes requested):
marking as requested changed to have a single commit. @ShyamsundarR please feel free to add a ready-to-merge label once you review.

matancarmeli7:
@Madhu-1  squashed

Madhu-1 (dismissed):

matancarmeli7:
@Madhu-1 I don't have permission to merge my changes
image

Madhu-1:
> @Madhu-1 I don't have permission to merge my changes
> image

it will get auto-merge when PR is approved by 2 developers (with write   permission)

matancarmeli7:
@Madhu-1 ok thanks

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+		Group:   replicationv1alpha1.GroupVersion.Group,
+		Kind:    resourceName,
+		Version: replicationv1alpha1.GroupVersion.Version,
+	})
+	for ! isVolumeReplicationResourceExists {
+		err := r.Client.List(context.TODO(), unstructuredResource)
+		if err != nil {
+			if meta.IsNoMatchError(err){
+				logger.Info("Resource", resourceName, "does not exist")
+				time.Sleep(5 * time.Second)
```
iamniting, line 363: It should not be indefinite IMO, There should be some timeout. @Madhu-1 What do you think?

iamniting (changes requested):

controllers/volumereplication_controller.go
Madhu-1: adding timeout doesn't solve the purpose of this change. we need to keep it infinite for loop.

iamniting (approved):

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+	err = r.waitForVolumeReplicationResource(logger, volumeReplication)
+	if err != nil {
+		logger.Error(err, "Failed to watch VolumeReplication")
+		return err
+	}
+	return nil
+}
+
+func (r *VolumeReplicationReconciler) waitForVolumeReplicationResource(logger logr.Logger, resourceName string) error {
+
```
sp98, line 350: remove new line here.

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
 		WithEventFilter(pred).Complete(r)
 }
 
+func (r *VolumeReplicationReconciler) waitForCrds() error {
+
```
sp98, line 332: remove new line here.

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+func (r *VolumeReplicationReconciler) waitForVolumeReplicationResource(logger logr.Logger, resourceName string) error {
+
+	isVolumeReplicationResourceExists := false
+	unstructuredResource := &unstructured.UnstructuredList{}
+	unstructuredResource.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   replicationv1alpha1.GroupVersion.Group,
+		Kind:    resourceName,
+		Version: replicationv1alpha1.GroupVersion.Version,
+	})
+	for ! isVolumeReplicationResourceExists {
```
sp98, line 358:
```suggestion
	for !isVolumeReplicationResourceExists {
```

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+
+	err = r.waitForVolumeReplicationResource(logger, volumeReplication)
+	if err != nil {
+		logger.Error(err, "Failed to watch VolumeReplication")
+		return err
+	}
+	return nil
+}
+
+func (r *VolumeReplicationReconciler) waitForVolumeReplicationResource(logger logr.Logger, resourceName string) error {
```
sp98, line 349: don't like the idea of passing logger as function argument.  We should explore other options if possible.

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+
+	isVolumeReplicationResourceExists := false
```
sp98, line 351: Suggesting to change this name. Doesn't sound correct. 
Suggestions -  `isResourcePresent`    `resourceExists`  `resourcePresent`.

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+		Version: replicationv1alpha1.GroupVersion.Version,
+	})
+	for ! isVolumeReplicationResourceExists {
+		err := r.Client.List(context.TODO(), unstructuredResource)
+		if err != nil {
+			if meta.IsNoMatchError(err){
+				logger.Info("Resource", resourceName, "does not exist")
+				time.Sleep(5 * time.Second)
+			} else {
+				logger.Error(err, "Got an unexpected error while waiting for", "Resource", resourceName)
```
sp98, line 365:
```suggestion
				logger.Error(err, "got an unexpected error while waiting for resource", "Resource", resourceName)
```

lets start with small case for consistency with log messages.

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
 		WithEventFilter(pred).Complete(r)
 }
 
+func (r *VolumeReplicationReconciler) waitForCrds() error {
+
+	logger := r.Log.WithName("checkingDependencies")
+
+	err := r.waitForVolumeReplicationResource(logger, volumeReplicationClass)
+	if err != nil {
+		logger.Error(err, "Failed to watch VolumeReplicationClass")
```
sp98, line 337:
```suggestion
		logger.Error(err, "failed to wait for VolumeReplicationClass CRD")
```

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,50 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+		return err
+	}
+
+	err = r.waitForVolumeReplicationResource(logger, volumeReplication)
+	if err != nil {
+		logger.Error(err, "Failed to watch VolumeReplication")
```
sp98, line 343:
```suggestion
		logger.Error(err, "failed to wait for VolumeReplication CRD ")
```

controllers/volumereplication_controller.go
matancarmeli7: Those names do not reflect what this function does.
This function waits until the resource is showing up, it doesn't tell if it exists or not.

controllers/volumereplication_controller.go
```diff
@@ -297,6 +302,12 @@ func (r *VolumeReplicationReconciler) updateReplicationStatus(instance *replicat
 // SetupWithManager sets up the controller with the Manager.
 func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.DriverConfig) error {
```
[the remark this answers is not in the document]
matancarmeli7: this is a line that I didn't add
https://github.com/csi-addons/volume-replication-operator/blob/2f60bc0065078c1b7ac7c2f99bb5f24e1d6c73c3/controllers/volumereplication_controller.go#L299

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
matancarmeli7: why? we do it before in multiple functions? and if you remove it from there, how do you think we should do it differently? creating the logger in the function? I am ok with passing the logger

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
sp98: since the method name is `waitForVolumeReplicationResource`,  it should be ok to remove the word `VolumeReplication`  from the variable name. Hence the above suggestions.

controllers/volumereplication_controller.go
sp98: you are right. Its done in other places as well. Ignore my above comment.

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,48 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+	unstructuredResource.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   replicationv1alpha1.GroupVersion.Group,
+		Kind:    resourceName,
+		Version: replicationv1alpha1.GroupVersion.Version,
+	})
+	for !isVolumeReplicationResourceExists {
+		err := r.Client.List(context.TODO(), unstructuredResource)
+		if err != nil {
+			if meta.IsNoMatchError(err){
+				logger.Info("Resource", resourceName, "does not exist")
```
sp98, line 360:
```suggestion
				logger.Info("resource does not exist", "Resource",  resourceName )
```

sp98 (commented):
one last nit.

controllers/volumereplication_controller.go
matancarmeli7: done

controllers/volumereplication_controller.go
matancarmeli7: done

Madhu-1 (approved):

controllers/volumereplication_controller.go
```diff
@@ -317,6 +328,48 @@ func (r *VolumeReplicationReconciler) SetupWithManager(mgr ctrl.Manager, cfg *co
...
+		Kind:    resourceName,
+		Version: replicationv1alpha1.GroupVersion.Version,
+	})
+	for !isResourceExists {
+		err := r.Client.List(context.TODO(), unstructuredResource)
+		if err != nil {
+			if meta.IsNoMatchError(err){
+				logger.Info("resource does not exist", "Resource",  resourceName)
+				time.Sleep(5 * time.Second)
+			} else {
```
sp98, line 362: We could remove the both the else statement in line 362 and 366 to reduce the nesting. 

```
 if err != nil {
      if meta.IsNoMatchError(err){
	  logger.Info("resource does not exist", "Resource",  resourceName)
	  time.Sleep(5 * time.Second)
          return nil
        } 
     logger.Error(err, "got an unexpected error while waiting for resource", "Resource", resourceName)
     return err
  } 
   isResourceExists = true
```

sp98 (changes requested):

controllers/volumereplication_controller.go
Madhu-1: if the error is noMatch we need to continue not to return nil. Maybe we can enhance that in the next PR. I can send one.

matancarmeli7: @sp98 we need them since we don't want to stop when the resource does not exists and only exit when there is a failure that doesn't relate to resource existence or that if the resource exists, otherwise we will want to stay in the loop

sp98 (approved):

PR #112, opened 2021-09-07
replication: update conditions when resync is done

### LINKED ISSUE #101: VolumeReplication condition "Degraded" never moves to "False" when recovering from a split brain
Test case:
- Create PVC in cluster1 (using rbd as the provisioner, NOTE: Assume all
  this is done on 2 minikube instances with rook-ceph and required
  VolRepClass and RBD)
- Create VR pointing to PVC above and state as "Primary" (in cluster1)
  - Ensure image on both clusters in rbd are present (using toolbox)
- Copy PV backing PVC on cluster1 to cluster2
- Create the same PVC on cluster2
  - Ensure it is bound to the PV
- Create VR on cluster2, pointing to the PVC above, and state as "Primary"
  - This is where the image goes into error and will further cause a split-brain
    --- Till here things are fine ---
- Update VR on cluster1 to "Secondary"
- Wait for VR conditions to reach the state (Completed:True; Resyncing:True; Degraded:False, for the current generation number (usually 2 in this test case))
  - VolRepl does not change "Degraded" to False at all, it remains "True"
  - Checked backend ceph image mirror status (or image info) on cluster1 and image reaches "state:       up+replaying"

Questions:
- Should the Condition `Degraded` move to `False` or is that an incorrect assumption to check if an image is a healthy `Secondary`?
- If the above is an incorrect assumption then the conditions to check that an image has re-established replication would just be `Completed: True` and `Resyncing: True`?

Other code observations:

The VolumeReplication reconciler does the following broadly when reconciling as Secondary:
- First loop iteration when marked as Secondary
  - markVolumeAsSecondary -> setDemotedCondition (Completed: True; Degraded: False; Resyncing: False)
  - updateReplicationStatus (orthogonal observation: seems redundant, as we just flip and set degraded to True, and update status again)
  - setOnlyDegradedCondition
  - updateReplicationStatus
- Second loop iteration when marked Secondary (or already Secondary)
  - markVolumeAsSecondary
  - updateReplicationStatus (orthogonal observation: again redundant, we should possibly update status once in the end per-return)
  - resyncVolume (conditionally returns requeueForResync id ready is not True in the response to a resync request)
  - updateReplicationStatus

ceph-csi ResyncVolume response:
- The ResyncVolume responds ready as [false](https://github.com/ceph/ceph-csi/blob/bb60173a9839b155d67b3c6c69d94a2973022201/internal/rbd/replicationcontrollerserver.go#L629) till [this](https://github.com/ceph/ceph-csi/blob/bb60173a9839b155d67b3c6c69d94a2973022201/internal/rbd/replicationcontrollerserver.go#L605-L607) condition is met

The question hence is, when resync is triggered for a split brain,  the assumption is that the conditions that should be met are: `Completed:True;
    Resyncing:True; Degraded:False`, is correct or not?
If not, then what are the conditions to wait for and when are they updated?

Madhu-1:
updating the conditions as below once the resync
is completed and volume can be marked ready to use

```
status:
    conditions:
    - lastTransitionTime: "2021-09-07T10:23:13Z"
      message: ""
      observedGeneration: 4
      reason: Demoted
      status: "True"
      type: Completed
    - lastTransitionTime: "2021-09-07T10:27:19Z"
      message: ""
      observedGeneration: 4
      reason: Healthy
      status: "False"
      type: Degraded
    - lastTransitionTime: "2021-09-07T10:27:19Z"
      message: ""
      observedGeneration: 4
      reason: NotResyncing
      status: "False"
      type: Resyncing
    lastCompletionTime: "2021-09-07T10:27:19Z"
    lastStartTime: "2021-09-07T09:38:43Z"
    message: volume is marked secondary
    observedGeneration: 4
    state: Secondary
```

The VR status condition will be

* Type completed status be True
* Type Degraded status will be False
* Type Resyncing status will be False

Fixes #101 

### DIFF
--- a/controllers/status.go
+++ b/controllers/status.go
@@ -32,6 +32,7 @@ const (
 	Success         = "Success"
 	Promoted        = "Promoted"
 	Demoted         = "Demoted"
+	Resync          = "Resync"
 	FailedToPromote = "FailedToPromote"
 	FailedToDemote  = "FailedToDemote"
 	Error           = "Error"
@@ -177,6 +178,28 @@ func setDegradedCondition(conditions *[]metav1.Condition, observedGeneration int
 		Reason:             ResyncTriggered,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionTrue,
+	})
+}
+
+// sets conditions when volume resync is completed and volume is ready to use
+func setVolumeReadyCondition(conditions *[]metav1.Condition, conditionReason string, observedGeneration int64) {
+	setStatusCondition(conditions, metav1.Condition{
+		Type:               ConditionCompleted,
+		Reason:             conditionReason,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionTrue,
+	})
+	setStatusCondition(conditions, metav1.Condition{
+		Type:               ConditionDegraded,
+		Reason:             Healthy,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionFalse,
+	})
+	setStatusCondition(conditions, metav1.Condition{
+		Type:               ConditionResyncing,
+		Reason:             NotResyncing,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionFalse,
 	})
 }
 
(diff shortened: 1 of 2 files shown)


Madhu-1:
Adding DNM until we finalize the status changes

Madhu-1:
Closing this one as its no longer required.

PR #113, opened 2021-09-14
Update conditions to reflect degraded and resyncing

ShyamsundarR:
The condition updates are when processing as secondary

- Resyncing
  - Is false on any errors
  - Is false if degraded is false
  - Is true, if resync is successful and volume is not
    yet "ready" to use
- Degraded
  - Is true on errors
  - Is true UNTIL resync returns "ready" response
  - Is false ONLY when resync returns "ready"

As Degraded moves to false, resyncing will move to
false as well, other than cases where there are
errors, or resync status is not yet determined. Resync
status is for example undetermined if we issued a
demote volume request, from markVolAsSecondary.

Resyncing: denotes if resync is active
Degraded: denotes local secondary contents are not
in full sync with remote, and that happens only when
remote is marked Secondary and local is also Secondary
and the final sync is complete.

Hence, Resyncing is false when degraded is false, as
there is no more data to sync.

Completed is always the initial condition to check before
looking at other conditions, to ensure operation was a
success. IOW, Completed being false, currently negates
any value in the other 2 conditions, and should typically
be Resync: false and Degraded: true.

Fixes: #101

Signed-off-by: Shyamsundar Ranganathan <srangana@redhat.com>

### DIFF
--- a/controllers/status.go
+++ b/controllers/status.go
@@ -86,13 +86,19 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
 	})
 }
 
-// sets conditions when volume is demoted and requeued for resync
-func setOnlyDegradedCondition(conditions *[]metav1.Condition, observedGeneration int64) {
+// sets conditions when volume is demoted and ready to use (resync completed)
+func setNotDegradedCondition(conditions *[]metav1.Condition, observedGeneration int64) {
 	setStatusCondition(conditions, metav1.Condition{
 		Type:               ConditionDegraded,
-		Reason:             VolumeDegraded,
+		Reason:             Healthy,
 		ObservedGeneration: observedGeneration,
-		Status:             metav1.ConditionTrue,
+		Status:             metav1.ConditionFalse,
+	})
+	setStatusCondition(conditions, metav1.Condition{
+		Type:               ConditionResyncing,
+		Reason:             NotResyncing,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionFalse,
 	})
 }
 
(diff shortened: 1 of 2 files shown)


Madhu-1:
Resyncing
Is false on any errors
Is false UNTIL, resync call returns a success
Is false if degraded is false

2nd statement `Is True once Resync call returns without error` for better readability?

controllers/status.go
```diff
@@ -86,13 +86,19 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
 	})
 }
 
-// sets conditions when volume is demoted and requeued for resync
-func setOnlyDegradedCondition(conditions *[]metav1.Condition, observedGeneration int64) {
+// sets conditions when volume is demoted and requeued for resync without errors
```
Madhu-1, line 89: sets conditions when volume is demoted, resync is completed and volume is ready to use?

ShyamsundarR:
@Madhu-1 updated addressing the current comments.

Madhu-1 (approved):

sp98 (approved):

controllers/volumereplication_controller.go
```diff
@@ -216,12 +216,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		if instance.Status.State != replicationv1alpha1.SecondaryState {
 			replicationErr = r.markVolumeAsSecondary(instance, logger, volumeHandle, parameters, secret)
 			if replicationErr == nil {
-				err := r.updateReplicationStatus(instance, logger, getReplicationState(instance), "volume is marked secondary")
```
oriyarde, line 219: how will the status state ever change from Primary to Secondary?
is this covered in any test? because after this change, the VR status state doesn't move to Secondary in our tests, but stuck in a `"volume is not ready to use"` loop

ShyamsundarR: The call to `updateReplicationStatus` is done after logging, so if there is an issue, it should not be at this point. The status.state is updated as part of `updateReplicationStatus` and that is called in the new code at line 220, right after `markVolumeAsSecondary` returns no errors.

ShyamsundarR: @Madhu-1 is testing this out now, and seeing the issue and debugging the same.

Madhu-1: Fix for it https://github.com/csi-addons/volume-replication-operator/pull/116

oriyarde: thanks

PR #114, opened 2021-09-14
add creation for multiarch images

matancarmeli7:


Makefile
```diff
@@ -32,6 +32,7 @@ BUNDLE_IMG ?= controller-bundle:$(VERSION)
 IMG_NAME ?= quay.io/csiaddons/volumereplication-operator
 IMG_TAG ?= latest
 IMG=${IMG_NAME}:${IMG_TAG}
+image_arch_list=linux/amd64,linux/ppc64le,linux/s390x
```
Madhu-1, line 35: make this a variable like IMG_NAME not everyone is interested in building multi-arch images locally for testing.

Makefile
```diff
@@ -94,11 +95,11 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build -t ${IMG} .
+	docker buildx build --tag ${IMG} --platform ${image_arch_list} .
```
Madhu-1, line 98: not everyone has buildx installed. maybe we need to have different makefile commands to build and push images with buildx

Makefile
```diff
@@ -94,11 +95,11 @@ generate: controller-gen
...
 
 # Push the docker image
-docker-push:
-	docker push ${IMG}
+docker-build-and-push:
```
Madhu-1, line 101: rename it to `dockerx-build-and-push-multi-arch`

matancarmeli7: done

Makefile
matancarmeli7: done

Makefile
matancarmeli7: I added an option to choose if to build multi-arch or not, so if someone will build multi-arch images he will use it, and if someone will choose the one with only x86 won't use it, so I think it can be good that way after the change

Madhu-1: the only problem is that now with multi-arch the image building takes more than expected. just want to avoid that

.github/workflows/docker-publish.yaml
```diff
@@ -25,6 +25,13 @@ jobs:
     - name: Set env
       # github tag will be like refs/tags/v0.1.0 :10 removes the first 10 characters
       run: echo RELEASE_VERSION=$(echo ${GITHUB_REF:10}) >> $GITHUB_ENV
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+      
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1.3.0
```
Madhu-1, line 34: use latest `1.6.0` version here and in  other place also

Madhu-1 (commented):
small nit

Makefile
Madhu-1: Discard this one.

.github/workflows/docker-publish.yaml
matancarmeli7: done

ShyamsundarR (approved):

PR #127, opened 2021-12-01
replication: add grpc message to CR status

Madhu-1:
instead of adding both the GRPC error code and the message to CR status just update the CR with the GRPC error message for better readability.

* Before Fix

```yaml
lastStartTime: "2021-11-30T06:46:52Z"
message: 'rpc error: code = Internal desc = failed to get local status: rbd: ret=-2,
No such file or directory'
observedGeneration: 2
state: Unknown
```
* After Fix

```yaml
lastStartTime: "2021-11-30T06:46:52Z"
message: 'failed to get local status: rbd: ret=-2, No such file or directory'
observedGeneration: 2
state: Unknown
```

### DIFF
--- a/controllers/replication/replication.go
+++ b/controllers/replication/replication.go
@@ -120,3 +120,15 @@ func (r *ReplicationResponse) HasKnownGRPCError(knownErrors []codes.Code) bool {
 
 	return false
 }
+
+// GetMessageFromError returns the message from the error.
+func GetMessageFromError(err error) string {
+	s, ok := status.FromError(err)
+	if !ok {
+		// This is not gRPC error. The operation must have failed before gRPC
+		// method was called, otherwise we would get gRPC error.
+		return err.Error()
+	}
+
+	return s.Message()
+}
--- a/controllers/replication/replication_test.go
+++ b/controllers/replication/replication_test.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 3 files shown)


controllers/replication/replication.go
```diff
@@ -120,3 +120,15 @@ func (r *ReplicationResponse) HasKnownGRPCError(knownErrors []codes.Code) bool {
 
 	return false
 }
+
+// getMessageFromError returns the message from the error.
+func GetMessageFromError(err error) string {
+	s, ok := status.FromError(err)
```
iamniting, line 126: Do we want to check if the error is nil?

Madhu-1: internally it's handled. if err is nil empty string will be returned.

iamniting (dismissed):

controllers/replication/replication.go
```diff
@@ -120,3 +120,15 @@ func (r *ReplicationResponse) HasKnownGRPCError(knownErrors []codes.Code) bool {
 
 	return false
 }
+
+// getMessageFromError returns the message from the error.
```
sp98, line 124:
```suggestion
// GetMessageFromError returns the message from the error.
```

sp98 (changes requested):
Nit. Rest looks good to me.

sp98 (dismissed):

controllers/replication/replication.go
```diff
@@ -121,7 +121,7 @@ func (r *ReplicationResponse) HasKnownGRPCError(knownErrors []codes.Code) bool {
 	return false
 }
 
-// getMessageFromError returns the message from the error.
+// GetMessageFromError returns the message from the error.
```
iamniting, line 124: I think this change should be in the first commit itself.

iamniting (changes requested):

iamniting (approved):

PR #137, opened 2023-05-19
Fix bug 2207632: infinite loop due to adding VR finalizer during PVC deletion

BenamarMk:
Resolve issue with VR falling into an infinite loop when attempting to add the finalizer "replication.storage.openshift.io/pvc-protection" during PVC deletion.

Fixing bug [2207632](https://bugzilla.redhat.com/show_bug.cgi?id=2207632)

Relevant gchat conversion is found [here](https://mail.google.com/chat/u/0/#chat/space/AAAAqWkMm2s/-h4y2DRqCcM):

### DIFF
--- a/controllers/volumereplication_controller.go
+++ b/controllers/volumereplication_controller.go
@@ -188,10 +188,12 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 			return reconcile.Result{}, err
 		}
-		if err = r.addFinalizerToPVC(logger, pvc); err != nil {
-			logger.Error(err, "Failed to add PersistentVolumeClaim finalizer")
+		if pvc.GetDeletionTimestamp().IsZero() {
+			if err = r.addFinalizerToPVC(logger, pvc); err != nil {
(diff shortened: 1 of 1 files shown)


controllers/volumereplication_controller.go
```diff
@@ -188,10 +188,12 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 			return reconcile.Result{}, err
 		}
-		if err = r.addFinalizerToPVC(logger, pvc); err != nil {
-			logger.Error(err, "Failed to add PersistentVolumeClaim finalizer")
+		if pvc.GetDeletionTimestamp().IsZero() {
```
Madhu-1, line 191: if the PVC is getting deleted should we just return error at https://github.com/csi-addons/kubernetes-csi-addons/blob/main/controllers/replication.storage/volumereplication_controller.go#L200 and let the next reconcile fail?

Madhu-1 (commented):
@BenamarMk we have deprecated the volume replication operator and migrated to kubernetes-csi addons, feel free to post the same change at https://github.com/csi-addons/kubernetes-csi-addons/blob/main/controllers/replication.storage/volumereplication_controller.go

